PR #17, opened 2021-03-10
Feature/update user syncdir

vilenarios:


vilenarios:
Added methods to support changing the user's sync folder path.  This will update the database, and rename the old folder to the new folder.

This could be improved by adding a proper move function instead of folder rename.

PR #19, opened 2021-04-09
Task/cleaning

matibat:
Changes

- Makes the const-to-function replacement
- Declares FolderWatchRef interface at files.ts
- Removes the bunch of imports at the end of some files


agsuy (approved):
Looking good. Linting seems good.
Merging

PR #22, opened 2021-05-19
Refactor/create classes and constructors for types

jdaev:
PR changes the new type interfaces to classes.
Classes have constructors and a From method to initialize with only default variables.
Added lastModifiedDate to types.
Fixed errors in gql.ts.


agsuy (approved):
Does build and all changes are whiting scope.

Merging

PR #23, opened 2021-05-19
task: split arweave into public and private

jdaev:
Moves arweave private functions into its own file
Added a constants.ts file.


agsuy (approved):
Did build and it is within scope.

Merging

PR #37, opened 2021-05-23
Query class

matibat:
### Changes

- Cleans code & refactors types, typeguards and type conditionals;
- Has an incomplete transaction-like structure to hold the queried values;
- Has a parameters array for setting the response parameters
- Auto-handles the nextPage property
- Auto-manages the min block value related to the lastBlock (if defined)
- Max-tries limit & backup server
- Implements methods to the raw transaction (query.getRaw()) and the FSEntity(query<The type>.get())


agsuy (changes requested):
Please rework _getSerializedTransactionData without the "IF"s

matibat:
Check it now

matibat:
Now:
- synced with dev

matibat:
The changes on types (implementing classes) now sets the properties' defaults to invalid (= 'invalid').
This makes us have cleaner constructors, just used for calling `Object.assign(this, args)` to set the initial values. This is a generic approach intended to work with any derivated class.

matibat:
Changes:

- Most of all queries replaced. Just missing getFileMetaDataFromTx(), which might be deleted after a better implementation of getAllMyDataFileTxs() to make it return directly the entities (already done, just need to be implemented)
- Private queries won't (yet) decrypt any data. **Let's see if we create a method for decrypting those to respect the SRP**
- All broken methods/types are at toDaemon/ folder

jdaev (approved):
Looks good to me

matibat:
Now all query methods that doesn't require decryption works well at CLI

matibat:
Now all query methods that doesn't require decryption works well at CLI

jdaev (approved):
Looks good

arielmelendez (changes requested):
Let's hold off on merging this. It's far too large for most of the team to comprehend. Let's talk more about that and the possibilities for moving forward later today. Thanks!

PR #47, opened 2021-06-01
Task/migrate db to loki

jdaev:
Adds loki based database classes and functions based on the following pattern:
Untitled Diagram


fedellen:
Closing this one, Core has been migrated and does not currently use a db

PR #48, opened 2021-06-03
RM-297: Use nvm to manage node engine version

arielmelendez:
• Using node 14.7.0 since its the most recent LTS version
• Updated README with nvm info


README.md
```diff
@@ -1,14 +1,24 @@
...
+  to enable git hooks for this local repository. Without doing so, you risk committing non-compliant code to the repository.
+</li>
+<li>Install all node package dependencies by running `yarn install --check-cache`</li>
+</ol>
 
-You will need to run `yarn husky install` to enable hooks locally on your repo
+## Building the Library
+Simply run `yarn build`. This will clean the project and compile the TypeScript library.
 
 ### VSCode/VSCodium integration:
```
arielmelendez, line 23: @agsuy , is it strictly necessary for yarn install information to be tucked under the IDE integration section? It seems redundant now that I've hoisted it up to the "## Building the Library" section.

jdaev (approved):
Looks good to me.

README.md
agsuy: I guess it deserves a small sub section after installing now but before the building one.

Suggested deps could use some formatting too I guess

agsuy: Also to install you could just use yarn install,

Per https://yarnpkg.com/cli/install --check-cache which "Always refetch the packages and ensure that their checksums are consistent" its a security measure needed for external PR.

matibat (approved):

README.md
arielmelendez: Ok we can improve that soon. Want to get the nvm stuff in first. Thanks for the feedback!

PR #49, opened 2021-06-09
feat: Add unit test for formatBytes function RM-301

fedellen:
This should handle all of the cases returned by `formatBytes` and hooks up the existing testing environment with Mocha and Chai.

### DIFF
--- a/src/common.test.ts
+++ b/src/common.test.ts
@@ -1,0 +1,34 @@
+import { expect } from 'chai';
+import { formatBytes } from './common';
+
+type byteTypes = 'Bytes' | 'KB' | 'MB' | 'GB';
+
+function formatBytesTestHelper(returnedString: string, type: byteTypes) {
+	const stringArray = returnedString.split(' ');
+
+	// Has correct string value, ex: 'Bytes'
+	expect(stringArray[1], `Incorrect string value: ${stringArray[1]}, expected: ${type}`).to.equal(type);
+
+	if (stringArray[0].includes('.')) {
+		// Byte value has been limited to 3 decimal places (thousandths), ex: 1743.432 KB
+		const decimalValue = stringArray[0].split('.')[1];
+		expect(decimalValue.length, `Decimal value is too long: "${stringArray[0]}"`).to.be.lessThan(4);
+	}
+}
+
+describe('The formatBytes function correctly returns', () => {
+	it('Bytes', () => {
+		// Very small integer for testing values without decimals
+		formatBytesTestHelper(formatBytes(85), 'Bytes');
+	});
+	it('KB', () => {
+		formatBytesTestHelper(formatBytes(8537), 'KB');
+	});
+	it('MB', () => {
+		formatBytesTestHelper(formatBytes(1262143), 'MB');
+	});
+	it('GB', () => {
+		// Very long integer for testing `.toFixed()` decimal places
+		formatBytesTestHelper(formatBytes(34737418246534), 'GB');
+	});
+});
--- a/src/common.test.ts
+++ b/src/common.test.ts
@@ -1,26 +1,53 @@
 import { expect } from 'chai';
 import { formatBytes } from './common';
 
-// Cases for rounding
-// Cases for max/min int
-// Cases for TB represented as GB
+describe('The formatBytes function ', () => {
+	describe('returns the properly formatted byte count description ', () => {
+		const inputToExpectedOutputMap = new Map<number, string>([
+			// Bytes range 0-1023
+			[0, '0 Bytes'],
+			[1023, '1023 Bytes'],
+			// KB range 1024-1048575
+			[1024, '1.000 KB'],
+			[1048575, '1023.999 KB'],
+			// MB range 1048576-1073741823
+			[1048576, '1.000 MB'],
+			[1073741823, '1024.000 MB'],
+			// GB range 1073741824-9007199254740991
+			[1073741824, '1.000 GB'],
+			[Number.MAX_SAFE_INTEGER, '8388608.000 GB'] // 9007199254740991 / 1024 / 1024 / 1024
+		]);
 
-describe('The formatBytes function correctly returns', () => {
-	const inputToExpectedOutputMap = new Map<number, string>([
-		[-12, '-12 Bytes'],
-		[0, '0 Bytes'],
-		[1023.999, '1023.999 Bytes'],
-		[1024, '1.000 KB'],
-		[85, '85 Bytes'],
-		[8537, '8.337 KB'],
-		[1262143, '1.204 MB'],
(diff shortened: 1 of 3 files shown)


jdaev (approved):
Looks good

fedellen:
Latest commits also introduce several edge cases that currently work:

```TypeScript
// TODO?: Should these next three cases even happen?
it('works with Bytes represented as a decimal', () => {
	expect(formatBytes(1023.999)).to.equal('1023.999 Bytes');
});

it('works with Bytes represented as a negative integer', () => {
	expect(formatBytes(-12)).to.equal('-12 Bytes');
});

it('returns values as "1024.000 MB" instead of "1.000 GB" when rounded up', () => {
	// Max range on MB: 1073741823
	expect(formatBytes(1073741823)).to.equal('1024.000 MB');
});
```

src/common.test.ts
```diff
@@ -0,0 +1,53 @@
...
+		]);
+
+		inputToExpectedOutputMap.forEach((expectedOutput, input) => {
+			it(`'${expectedOutput}' when the byte count input is ${input}`, () => {
+				expect(formatBytes(input)).to.equal(expectedOutput);
+			});
+		});
+	});
+
+	it('successfully rounds up and down', () => {
```
arielmelendez, line 28: Might want to experiment with rounding behavior of negative numbers. I doubt we've spec'ed how we EXPECT that to work, but we might as well test for the behavior. Then leave a note in the test that we're only testing it so we can detect any potential change in behavior in the future, NOT because we've intended for it to work that way.

src/common.test.ts
```diff
@@ -0,0 +1,53 @@
...
+		// 1074341824 / 1024 / 1024 / 1024 = 1.000465661 GB rounds down to 1.000 GB
+		expect(formatBytes(1074241824)).to.equal('1.000 GB');
+	});
+
+	it('represents TB sizes as GB', () => {
+		// 34737418246534 / 1024 / 1024 / 1024 / 1024 = 31.593497848 TB, returns as 32351.742 GB
+		expect(formatBytes(34737418246534)).to.equal('32351.742 GB');
+	});
+
+	// TODO?: Should these next three cases even happen?
```
arielmelendez, line 40: This is a great question. Any opinions @vilenarios @jdaev @matibat @agsuy ?

src/common.test.ts
fedellen: So the answer is that negative byte counts don't round or convert, they are another edge case:

```TypeScript
it('does not round or convert large negative byte counts', () => {
  expect(formatBytes(-57138495792)).to.equal('-57138495792 Bytes');
});
```

arielmelendez: YIKES. Perhaps we should consider throwing an exception if we encounter a negative number. I'd like to know more about how the function is used to know how risky that might be to the application in its current state.

src/common.test.ts
```diff
@@ -0,0 +1,58 @@
...
+
+	it('works with byte count represented as a negative integer', () => {
+		expect(formatBytes(-12)).to.equal('-12 Bytes');
+	});
+
+	it('does not round or convert large negative byte counts', () => {
+		expect(formatBytes(-57138495792)).to.equal('-57138495792 Bytes');
+	});
+
+	it('returns values as "1024.000 MB" instead of "1.000 GB" when rounded up', () => {
```
arielmelendez, line 54: This is arguably a bug!

src/common.test.ts
jdaev: The negative bytes are surely a bug. The only places I believe formatBytes is used should be in displaying file size and that isn't ever supposed to be negative.

jdaev: I guess negative and other unexpected valued could also happen during incorrect type conversion.

agsuy: Maybe the real question here is IF using formatBytes to display file size is the best approach.

agsuy:
> This should handle all of the cases returned by `formatBytes` and hooks up the existing testing environment with Mocha and Chai.

Maybe it is out of scope but maybe listing which are "all the cases" returned by formatBytes would be helpful.

Generally my approach would be to:

- Check intended use 
- Constrain I/O to needed/expected ones 
- List all relevant test cases (TC) with input and expected output 
- Review those so we ensure adequate/proper coverage
- Write tests

A good clue of a flaw design-wise would be when you need too many TC to cover a function that should be doing "just one thing"

A good measure against that would be using your constraints from point 2 to apply some input sensitization reducing your valid inputs to those that you need.

fedellen:
> Maybe it is out of scope but maybe listing which are "all the cases" returned by formatBytes would be helpful.

I agree that I should document the main cases better in the test file. This is also offered in the output when running the tests:
Screenshot from 2021-06-10 10-45-11

The main three cases being: 
- `successfully rounds up and down`
- `represents TB sizes as GB`
- `returns the properly formatted byte count description` (checks all of the byte count ranges (Bytes-KB KB-MB MB-GB))

The remaining four cases are simply there to point out possible flaws in the function. One of these cases did bring up something that could be considered a bug, as mentioned above by Ariel:

```TypeScript
it('returns values as "1024.000 MB" instead of "1.000 GB" when rounded up', () => {
  // Max range on MB: 1073741823
  expect(formatBytes(1073741823)).to.equal('1024.000 MB');
});
```
While accurate, I don't think `1024.000 MB` is the intended result.

---

> A good clue of a flaw design-wise would be when you need too many TC to cover a function that should be doing "just one thing"

Absolutely, ideally I think this function would be reworked so extra test cases would not be required, and the function would not even accept negative integers or byte counts as decimals. 

But this case was mostly an exercise to find something very simple to test. The function does not appear in the CLI repo, and is currently only used once in `ardrive-core-js` to prepare an upload batch:

https://github.com/ardriveapp/ardrive-core-js/blob/78afef274bef6f7fc9488374edee69448472840d/src/node.ts#L170-L171

https://github.com/ardriveapp/ardrive-core-js/blob/78afef274bef6f7fc9488374edee69448472840d/src/node.ts#L233-L238

Since it's hardly being used the question also becomes, is converting bytes to a readable/standard format something that we will want in the core to be used in all of our apps? Probably, it seems common enough since we're offering data storage. We do not however, offer data deletion. So maybe negative integers are out of scope entirely.

fedellen:
Added some info at the top in the latest push:

```TypeScript
```

arielmelendez:
I agree with everything @agsuy and @fedellen wrote re: ideals and path forward. I think we are good to go and can do an exercise as a team to improve this function. I have some good "Designing for Testability" insights that I'd like to share. Now that we have the test cases in place, we'll be able to refactor with confidence.

arielmelendez (approved):

agsuy (approved):

PR #50, opened 2021-06-10
RM-307 GitHub action yarn test and build

arielmelendez:
Introduces a github action to test and build the library.

### DIFF
--- a/.github/workflows/nodejs-ci.yml
+++ b/.github/workflows/nodejs-ci.yml
@@ -13,7 +13,7 @@ jobs:
         with:
           node-version: '14.x'
       - name: Install dependencies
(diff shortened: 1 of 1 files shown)


.github/workflows/nodejs-ci.yml
```diff
@@ -0,0 +1,20 @@
...
+
+on: [ push, pull_request ]
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js 14.x
```
agsuy, line 11: Should not we be using v14.17.0 here based on core .nvmrc?

PR #52, opened 2021-06-10
Meaningful type guards

matibat:


src/types/type_guards.ts
```diff
@@ -1,5 +1,35 @@
-export type cipherType = 'aes-gcm-256' | 'AES256-GCM' | 'invalid' | string;
-export type entityType = 'drive' | 'file' | 'folder' | 'invalid' | string;
-export type drivePrivacy = 'private' | 'public' | 'invalid' | string;
-export type driveAuthMode = 'password' | 'invalid' | string;
-export type syncStatus = 0 | 1 | 2 | 3;
+export type emptyString = '';
+export const emptyString: emptyString = '';
```
matibat, line 2: Hmmm I think type_guards.ts is not the place for const declarations

Should it be moved to common maybe?
```suggestion
```

jdaev: I'd made a separate constants.ts file. Could add it there

fedellen: I'm not sure we need both a type declaration and a constant in this case. One of TypeScript's core features is to infer as many types as possible so we don't have to

```ts
export const emptyString = '';
```

Declaring any string with `const` will force the type checker to infer it as a template literal. Which results in `emptyString` only having a possible type of `''`

Usage would still have the same meaningful readability that you're looking for here:

```ts
const drive = {
// ...
cipherIV: emptyString,
cipher: emptyString
// ...
```

This isn't very critical, just something I thought about while reading through the diff

arielmelendez:
Please reintegrate with dev!

arielmelendez:
> Please reintegrate with dev!

Actually I'll take care of that for you.

src/common.ts
```diff
@@ -336,14 +337,13 @@ export async function createNewPublicDrive(login: string, driveName: string): Pr
 		appVersion: appVersion,
 		driveName,
 		rootFolderId,
-		cipher: '',
+		cipher: AES256_GCM,
```
arielmelendez, line 340: Why is this change included in this PR?

matibat:
Alright, guys. I just removed the "extra changes" I applied to this PR. Now it's changing just what needed, removed the unnecessary const definitions.

src/common.ts
matibat: Just removed that change. The constant was already defined in another file.
Also the value there should still be `''`.

src/common.ts
```diff
@@ -336,14 +337,13 @@ export async function createNewPublicDrive(login: string, driveName: string): Pr
 		appVersion: appVersion,
```
arielmelendez: This is as high as I can place this comment, but the documentation for this function and its name are misleading. 

• The documentation says that it actually creates the drive and adds it to the drive table. In reality, it just creates an ArFSDriveMetaData object instance with what we might consider to be "default" values for a public drive, setting the passed in "login" (whatever that is) and drive name along the way. It then prints a misleading log statement. And for reasons not clear to me, this is all packaged up as a Promise for use at a later time.

• Similar to the documentation, the name of the function implies that a new drive will actually be created. But in reality, all that's created is metadata which, by the time this function is done running and returning, isn't necessarily associated with any DB entity. Perhaps a better name might be "createPublicDriveMetadata". It removes the redundant "new" after "create" and actually implies the correct return type.

We don't have to make these changes now. But I thought about them because there are currently no unit, integration, contract, or end-to-end tests for this function. So I have no way to know whether the changes you've introduced (e.g. the encryption type) have broken something somewhere else. In thinking about what kind of test function we could write, I tried to dig in and understand the code. And that's what led me to these conclusions.

So what can we do? One thought: deprecate this function and create a new one that meets our needs for the future and add a proper unit test for that function. What do you think?

src/types/type_guards.ts
matibat: For now I'm removing that consts. It also simplifies the committed diff

src/common.ts
matibat: There is already an addDriveToDriveTable that does it separately.

We should add a task new for updating the comment and names of those functions.

src/gql.ts
```diff
@@ -1720,7 +1721,7 @@ export async function getSharedPublicDrive(driveId: string): Promise<types.ArFSD
 						drive.arFS = value;
 						break;
 					case 'Drive-Privacy':
-						drive.drivePrivacy = value;
+						drive.drivePrivacy = value as drivePrivacy;
```
arielmelendez, line 1724: Did we actually gain any type safety here? From what I understand, the `as` operator is pretty lax. See here:
https://thoughtbot.com/blog/the-trouble-with-typescript-enums

src/gql.ts
```diff
@@ -1800,7 +1801,7 @@ export async function getPrivateDriveRootFolderTxId(
 ): Promise<types.ArFSRootFolderMetaData> {
 	let rootFolderMetaData: types.ArFSRootFolderMetaData = {
 		metaDataTxId: '0',
-		cipher: '',
+		cipher: AES256_GCM,
```
arielmelendez: How can we know that this didn't break anything?

src/gql.ts
```diff
@@ -1840,7 +1841,7 @@ export async function getPrivateDriveRootFolderTxId(
 				const { value } = tag;
 				switch (key) {
 					case 'Cipher':
-						rootFolderMetaData.cipher = value;
+						rootFolderMetaData.cipher = value as cipherType;
```
arielmelendez, line 1844: Some comment here about not actually getting type safety guarantees here:
https://thoughtbot.com/blog/the-trouble-with-typescript-enums

src/public/arweave.ts
```diff
@@ -121,7 +121,9 @@ export async function prepareArFSDriveTransaction(
 		transaction.addTag('Content-Type', 'application/octet-stream');
 		transaction.addTag('Cipher', driveMetaData.cipher);
 		transaction.addTag('Cipher-IV', driveMetaData.cipherIV);
-		transaction.addTag('Drive-Auth-Mode', driveMetaData.driveAuthMode);
+		if (driveMetaData.driveAuthMode) {
```
arielmelendez, line 124: Should it be an error if there's no driveAuthMode? Also, do we need a type for the possible supported auth mode types?

src/types/type_guards.ts
arielmelendez: Are we really gaining clarity or conciseness by going from a canonical literal `''` to `emptyString`?

src/gql.ts
fedellen: I have read to avoid TypeScript enums in a couple books, but the TS docs themselves don't seem to discourage it's use. When I think of type guards I think of creating a function that proves to JavaScript that the type is correct at runtime:

```ts
const drivePrivacy = ['private', 'public', ''];
type drivePrivacy = typeof drivePrivacy[number];

function isDrivePrivacy(value: unknown): value is drivePrivacy {
  return typeof value === 'string' && drivePrivacy.includes(value);
}
```
The `is` keyword is the main driver of this type-guard method.

src/gql.ts
matibat: This is only for the compiler to do not throw that error for now (not enough type overlap on the cast of string to 'private' | 'public'). Please have in mind all the entire gql.ts files gets refactored by [the task #360 - Clean gql.ts](http://redmine.ardrive.io/issues/306)

src/public/arweave.ts
matibat: ### Should it be an error if there's no driveAuthMode?

I found in a comment on the code that the only supported value of driveAuthMode for now is: "password"; but it's set as an empty string by default.
So I made it to be nullable. Now driveAuthMode at any entity should be either "password" or undefined.

**Note** It shouldn't break a thing because in the case driveAuthMode was not "password", it was set as "" anyway.

### do we need a type for the possible supported auth mode types?

There already is a driveAuthMode type added by this PR. It is the string literal 'password'

PR #53, opened 2021-06-15
[TASK RM-303] Strict guards

matibat:
### Changes

- Removes the `string`, `""`, and `"invalid"` from the guard types;
- fixes all compiler errors involved in the change.

#### Note

I'm still using the `as` keyword, and also using union types with `""` to quick fix the compiler errors **without touching the algorithm**. That would be replaced in an upcoming PR for task RM-304.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -1,4 +1,4 @@
-import { cipherType } from './types/type_guards';
+import { CipherType } from './types/type_guards';
 
 export const prodAppUrl = 'https://app.ardrive.io';
 export const stagingAppUrl = 'https://staging.ardrive.io';
@@ -9,7 +9,7 @@ export const appName = 'ArDrive-Desktop';
 export const webAppName = 'ArDrive-Web';
 export const appVersion = '0.1.0';
 export const arFSVersion = '0.11';
-export const cipher: cipherType = 'AES256-GCM';
+export const cipher: CipherType = 'AES256-GCM';
 
 //Note: Just to easily copy paste later where it's needed
 //import {prodAppUrl,stagingAppUrl,gatewayURL,appName,webAppName,appVersion,arFSVersion,cipher} from './constants';
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -9,7 +9,7 @@ export const appName = 'ArDrive-Desktop';
 export const webAppName = 'ArDrive-Web';
 export const appVersion = '0.1.0';
 export const arFSVersion = '0.11';
-export const cipher: CipherType = 'AES256-GCM';
+export const defaultCipher: CipherType = 'AES256-GCM';
 
 //Note: Just to easily copy paste later where it's needed
 //import {prodAppUrl,stagingAppUrl,gatewayURL,appName,webAppName,appVersion,arFSVersion,cipher} from './constants';
--- a/src/types/base_Types.ts
+++ b/src/types/base_Types.ts
@@ -19,7 +19,7 @@ export interface UploadBatch {
 
 export interface ArFSRootFolderMetaData {
 	metaDataTxId: string;
-	cipher: guards.cipherType;
+	cipher: guards.CipherType | '';
 	cipherIV: string;
 }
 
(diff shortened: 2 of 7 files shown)


fedellen (approved):
This looks good to me. I'm not sure why Ariel's commits are showing up here, but they don't conflict in any way. The generic `UnionOfObjectPropertiesType` is a great idea for capturing all the object values.

I see the `as` casting here is just a result of narrowing the types. At some point we can guard those inputs as well, as I mentioned to you on slack: 

```ts
function isCorrectType<T>(value: unknown, object: T): value is T {
  return Object.values(object).includes(value as T)
}
```

matibat:
The validation of the inputs might be done after [the task #304](http://redmine.ardrive.io/issues/304) adds well-defined constructors with conditional arguments and default values, so we can validate it from the structure itself.

agsuy:
Do we have any documentation about the changes implemented here?

matibat:
Please note the all the `as` occurrences are wrapped at gql.ts, the file to be refactored

matibat:
> Do we have any documentation about the changes implemented here?

https://github.com/ardriveapp/docs/blob/main/Core/cleaning_gql_methods.md

src/types/base_Types.ts
```diff
@@ -38,14 +38,14 @@ export interface ArFSDriveMetaDataParameters {
 	appVersion: string;
 	driveName: string;
 	rootFolderId: string;
-	cipher: guards.cipherType;
+	cipher: guards.CipherType | '';
```
arielmelendez, line 41: Why do we need to support `| ''`? It seems like we're just creating another type here that uses empty string to mean something to some piece of business logic somewhere else. It's not clear to the reader what that purpose is. Codifying that as a type, itself, could be helpful. OR, perhaps, eliminating that union altogether, if possible, would be preferable.

src/constants.ts
```diff
@@ -9,7 +9,7 @@ export const appName = 'ArDrive-Desktop';
 export const webAppName = 'ArDrive-Web';
 export const appVersion = '0.1.0';
 export const arFSVersion = '0.11';
-export const cipher: cipherType = 'AES256-GCM';
+export const cipher: CipherType = 'AES256-GCM';
```
arielmelendez, line 12: Could this constant benefit from a name change, like `defaultCipher`? `cipher` seems like it could clash with lots of locally scoped variables.

jdaev: I added that at the time because I couldn't think of a better name. Changing to to something that makes more sense would be great. As for the empty strings, those are there so the legacy functions would still work while refactoring.

src/types/base_Types.ts
jdaev: It's for ensuring that the older legacy functions will still work.

matibat: That's what the footer **Note** of the description talks about: I left those unions and `as` keywords temporarily still in the code in a way it doesn't affect the algorithm, just the types.

Those files are refactored at #54 :)

src/constants.ts
matibat: Definitely!

src/types/base_Types.ts
agsuy: This might be related to the way Phil implemented entities in the original core. 

We need to get rid of those '' representations.

Those empty string are here and there like [here in base types](https://github.com/ardriveapp/ardrive-core-js/blob/14450462d59c794eb58ad86740a6b6abbc97292c/src/types/base_Types.ts#L111) or [here in arfs](https://github.com/ardriveapp/ardrive-core-js/blob/14450462d59c794eb58ad86740a6b6abbc97292c/src/types/arfs_Types.ts#L73) also field generator [some lines below](https://github.com/ardriveapp/ardrive-core-js/blob/14450462d59c794eb58ad86740a6b6abbc97292c/src/types/arfs_Types.ts#L166)

arielmelendez: Ok I will wait patiently for improvements in PR 54 😬

matibat:
Just renamed `cipher` to `defaultCipher`

arielmelendez (approved):

PR #54, opened 2021-06-16
[TASK RM-304] Refactor types to have a clean constructor for each entity

matibat:
### Changes

- Refactors entities types to
  - have all of them as a class;
  - have an interface per entity;
  - allow constructing a new entity with default values of parameters.
- Fixes all type issues after the refactor.


matibat:
The task: http://redmine.ardrive.io/issues/304
The diagrams: http://redmine.ardrive.io/issues/308

arielmelendez:
@matibat , could you please make the merge destination feature/strict-guards until that branch is merged to dev (at which point you can change the merge destination for this PR to dev). That will allow us to see the delta in the changes introduced by this PR. Thanks!

matibat:
Please note: here -as in task #303- I'm not touching the algorithm, but only the type definitions.
After the current code works with the new data structures we'd start the real magic on task #306

matibat:
A resume of the commits:

1. The first commit (3facf697a0c824c2922a6f2b3b26e313eb0d472c) adds the checking function for the specific types. Yet not calling any
2. The refactor itself happens at 226b60a5606d72005e5abffa96f08c829eff86c5 which edits `src/types/*.ts` files. To have a clear visualization of what's happening, there's a diagram in the subtask [RM-308](http://redmine.ardrive.io/issues/308); current origin/dev on the left, and the new definitions on the right.
3. Then the following one (b62ace7c1afd4d59169265dc39396c1971c52f48) fixes errors that raised as consequence.
4. The instantiation of objects changes for calling new() with constructors on 9ac9e79ec6025ee4b26be2197d34ecc592b39980, where the default values start taking effect on removing it the responsibility of defining ALL the values at once. Just define the known values when constructing, then fill it up at runtime.
5. & 6. And a few more non-interesting commits. A minor fix here: 4ea800487520832c653bd84804812f46cfae3b06 and a merge for syncing the base branch: 0d64619c1d05a9109e41b31ba73f4cee0e78352c

src/types/arfs_Types.ts
```diff
@@ -22,185 +39,164 @@ export interface Wallet {
...
+	// driveId?: string; // the unique drive identifier, created with uuidv4 https://www.npmjs.com/package/uuidv4 eg. 41800747-a852-4dc9-9078-6c20f85c0f3a
+	entityType: EntityType = entityTypeValues.FILE; // the type of ArFS entity this is.  this can only be set to "drive", "folder", "file"
+	name = ''; // user defined entity name, cannot be longer than 64 characters.  This is stored in the JSON file that is uploaded along with the drive/folder/file metadata transaction
+	syncStatus: SyncStatus = syncStatusValues.READY_TO_DOWNLOAD; // the status of this transaction.  0 = 'ready to download', 1 = 'ready to upload', 2 = 'getting mined', 3 = 'successfully uploaded'
+	txId = ''; // the arweave transaction id for this entity. 43 numbers/letters eg. 1xRhN90Mu5mEgyyrmnzKgZP0y3aK8AwSucwlCOAwsaI
+	unixTime = 0; // seconds since unix epoch, taken at the time of upload, 10 numbers eg. 1620068042
+
+	constructor(args: T) {
+		// super();
+		Object.assign(this, args);
```
matibat, line 116: The way I set the attribute values of all `ArFSEntity` instances (and derivated classes) is by directly applying the args attributes to `this`.

It may cause trouble if `T` does not extend IEntity. Eg.: if the IEntity argument is set dynamically when it comes from the internet

src/types/base_Types.ts
```diff
@@ -31,258 +47,138 @@ export interface ArDriveBundle {
...
-	metaDataTxId: string;
-	metaDataSyncStatus: number;
-	isLocal?: number;
+export class ArDriveUser implements IDriveUser {
+	login = '';
+	dataProtectionKey = '';
+	walletPrivateKey = '';
+	walletPublicKey = '';
+	syncFolderPath = '';
+	autoSyncApproval: YesNoInteger = 0;
```
fedellen, line 76: This type change is stopping the CLI from compiling. Pretty easy fix though, line 82 in `ardrive-cli`:

```diff
- export const promptForAutoSyncApproval = async (): Promise<number> => {
+ export const promptForAutoSyncApproval = async (): Promise<YesNoInteger> => {
```

src/types/type_guard_functions.ts
```diff
@@ -0,0 +1,24 @@
...
+export const isCorrectCipherType = isCorrectValueFactory.bind(this, cipherTypeValues);
+export const isCorrectEntityType = isCorrectValueFactory.bind(this, entityTypeValues);
+export const isCorrectContentType = isCorrectValueFactory.bind(this, contentTypeValues);
+export const isCorrectDrivePrivacy = isCorrectValueFactory.bind(this, drivePrivacyValues);
+export const isCorrectDriveAuthMode = isCorrectValueFactory.bind(this, driveAuthModeValues);
+export const isCorrectDriveSharing = isCorrectValueFactory.bind(this, driveSharingValues);
+export const isCorrectSyncStatus = isCorrectValueFactory.bind(this, syncStatusValues);
+export const isCorrectYesNoInteger = isCorrectValueFactory.bind(this, yesNoIntegerValues);
+
+function isCorrectValueFactory<T>(o: { [key: string]: T }, v: unknown): boolean {
```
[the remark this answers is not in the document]
matibat: Well, I tried to apply the cange by using function factories for less code repetition, but anyways the IDE didn't catch the `value is T` it's the same than I wrote `boolean` as returning type.

```ts
export const isCorrectYesNoInteger = (value: unknown): value is YesNoInteger =>
	isCorrectValue<YesNoInteger>(yesNoIntegerValues, value);

function isCorrectValue<T extends string | number>(o: { [key: string]: T }, v: unknown): v is T {
	const isCorrect = Object.values(o).includes(v as T);
	return isCorrect;
}
```

It's fixed by specifying the explicit type 

```ts
isCorrectYesNoInteger =  (value: unknown): value is YesNoInteger
```

===================

I also tried by using ReturnType<typeof theFunction> but it returns `boolean`

src/types/base_Types.ts
```diff
@@ -31,258 +36,107 @@ export interface ArDriveBundle {
...
-			arFS: '',
-			driveId,
-			driveSharing: 'shared',
-			drivePrivacy: 'public',
-			driveAuthMode: '',
-			metaDataTxId: '0',
-			metaDataSyncStatus: 0
-		});
+	constructor(protected readonly template: IDriveUser = {}) {
+		this.autoSyncApproval = Number(this.autoSyncApproval) as YesNoInteger;
```
fedellen, line 51: Getting an error from the CLI with this change: 

```shell
src/index.ts:15:6 - error TS2741: Property 'template' is missing in type '{ login: string; dataProtectionKey: string; walletPrivateKey: string; walletPublicKey: string; syncFolderPath: string; autoSyncApproval: 0; }' but required in type 'ArDriveUser'.
```

src/types/arfs_Types.ts
```diff
@@ -22,185 +41,144 @@ export interface Wallet {
...
+export type IPrivateFileFolderEntity = IFileFolderEntity<PrivateType>;
+
+export type IFileData<P extends DrivePrivacy> = Partial<ArFSFileData<P>>;
+
+export type IPublicFileEntity = IFileData<PublicType>;
+
+// export type IPrivateFileEntity = IFileData<PrivateType>;
+
+export abstract class ArFSEntity<T extends IEntity> implements IEntity {
+	[key: string]: unknown;
```
fedellen, line 84: The `[key: string]: unknown` changes here are a pretty large. They mean anything else can also be added on to the object, which could be accurate if the source is unsanitized.

I'm curious why we needed these added, or your thought process behind it. At the very least, I'd make a comment to document why such a wide type would be necessary in these cases.

matibat: Yes, that wouldn't be there. It was to solve a conflict on the test files by trying to index the keys of those classes. I'll have to take another approach and remove that from the signature.

src/types/base_Types.ts
matibat: That is because that's defining an object from the class type instead of calling the constructor. The way of having an actual instance is to pass that object as the argument of the constructor.
```ts
let user = new ardrive.ArDriveUser({}); // All values are the defaults, so no need to specify them
```
instead of
```ts
let user: ardrive.ArDriveUser = {
		login: '',
		dataProtectionKey: '',
		walletPrivateKey: '',
		walletPublicKey: '',
		syncFolderPath: '',
		autoSyncApproval: 0
	};
```

That private & readonly template object is used to set the initial (or default) values in the constructor.

====================

I'll add a PR at CLI to apply the new way of instantiation. Thanks for advising!

src/types/arfs_Types.ts
matibat: This is replaced with the safer approach:

- To make the single taken argument of the constructor be read-only and protected; the `template` property which is a partial version of the class.
- As that argument is set as a property of the class, it can be accessed at instantiation time and be used to optionally set the value it holds instead of the default.

No extra properties will be added. Any unset value is the same as its default.

src/types/base_Types.ts
matibat: Added the fix at [PR-9](https://github.com/ardriveapp/ardrive-cli/pull/9)

src/types/arfs_Types.ts
matibat: Alright, just removed that

src/types/base_Types.ts
matibat: I think it's also fixed there: [PR-9 at cli](https://github.com/ardriveapp/ardrive-cli/pull/9)

src/gql.ts
```diff
@@ -1721,7 +1562,7 @@ export async function getSharedPublicDrive(driveId: string): Promise<types.ArFSD
 						drive.arFS = value;
 						break;
 					case 'Drive-Privacy':
-						drive.drivePrivacy = value as DrivePrivacy;
+						// drive.drivePrivacy = value;
```
fedellen, line 1565: I know this file is intended to be heavily refactored in the upcoming task, and some of these gql functions aren't even being used right now -- but are these safe to leave commented out here?

matibat: More than safe: It is mandatory to do not assign some properties (as the drivePrivacy, contentType, or entityType) because these are now readonly.
If we are querying a private entity, we already know it's private.

fedellen:
I think this branch needs to reintegrate with dev to get the test-setup commits out of [PR-55](https://github.com/ardriveapp/ardrive-core-js/pull/54).

This PR is probably testing some stuff that should *just work* without testing, but I don't think there can be too many tests. I am curiously awaiting to read @arielmelendez's review on the tests and constructors.

Everything else seems good to go on my end. I don't have much personal experience using classes but I can follow whats happening and it results in cleaning up a lot of redundant code. 

We need to make sure Ariel and Phil know that the CLI will temporarily break until it can reintegrate with this update.

matibat:
Outdated. Closing...

PR #55, opened 2021-06-16
[TASK RM-306] Clean gql.ts

matibat:


matibat:
This is way outdated. Closing...

PR #56, opened 2021-06-22
PE-3: Setup supporting testing libraries

fedellen:
This PR sets up Sinon for mocking, Istanbul for code coverage, and power-assert as a powerful debugging tool. It also adds the configuration files necessary to get up and running with each library.

I wrote an `example.test.ts` file to show and test the functionality of each library. This file is also a placeholder for where the integration tests would eventually be. I wanted to make sure we could test files in that directory with this setup.

The process of using the testing environment is documented in the `README` for now.


arielmelendez (changes requested):
yarn cache files should not be in this pull request. They'll blow up our repo :)

https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored

arielmelendez (changes requested):
Please update the README with instructions for testing the codebase.

fedellen:
> yarn cache files should not be in this pull request. They'll blow up our repo :)
> 
> https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored

I almost asked about this one when they started showing up in my commits, but I did go to that exact link which is referenced in our `.gitignore` file. We have the same setup they recommend with "Zero-Installs":

https://github.com/ardriveapp/ardrive-core-js/blob/28403296465f1ba48affb336a60f4b5f1f2d37f3/.gitignore#L41-L50

If we're not using the Zero-Installs I can safely change that up.

> Please update the README with instructions for testing the codebase.

Sounds good! I'll add some more testing instructions to the README.md

arielmelendez:
> If we're not using the Zero-Installs I can safely change that up.
My understanding is that the team doesn't use zero-installs day-to-day. We only use it when accepting external PRs (which rarely happen if I'm not mistaken).

In thinking about external PRs, I'm not sure if that means we HAVE to use zero-installs all the time or not, specifically when we think about the impact on our CI system (i.e. CI would run for the PR and we'd have to think about potential impacts of the build container ingesting the potential hazardous external lib).

@agsuy , any insights?

agsuy:
> > If we're not using the Zero-Installs I can safely change that up.
> 
> My understanding is that the team doesn't use zero-installs day-to-day. We only use it when accepting external PRs (which rarely happen if I'm not mistaken).
> 
> In thinking about external PRs, I'm not sure if that means we HAVE to use zero-installs all the time or not, specifically when we think about the impact on our CI system (i.e. CI would run for the PR and we'd have to think about potential impacts of the build container ingesting the potential hazardous external lib).
> 
> @agsuy , any insights?

The main drive of using Yarn2 IS Zero Installs (use repo as soon as you clone it) so we all resolve the same packages across all envs and there is not local regeneration (yarn install) lowering overall entropy. https://yarnpkg.com/features/zero-installs

matibat (approved):
Seems good!

README.md
```diff
@@ -41,6 +42,157 @@ yarn install --check-cache
 
 for external PR
 
+## Testing the Library
```
arielmelendez, line 45: Might make sense to move this above the VS Code integration section.

README.md
```diff
@@ -41,6 +42,157 @@ yarn install --check-cache
...
+
+This library is setup for [Mocha] testing with [Chai] and [Sinon].
```
arielmelendez, line 47: Did you intend to add some URL links here for Mocha, Chai, and Sinon?

README.md
```diff
@@ -41,6 +42,157 @@ yarn install --check-cache
...
+To run all of the tests use:
+
+```shell
+yarn test
+```
+
+Or, to run a specific test use Mocha's [grep] command:
+
+```shell
+yarn test -g 'my specific unit test'
```
arielmelendez, line 58: Can you give some more descriptive examples here? E.g. specific test names, specific test classes/files, specific folders?

README.md
arielmelendez: Same for other links below.

README.md
```diff
@@ -41,6 +42,157 @@ yarn install --check-cache
...
+```
+
+### Coverage
+
+[Istanbul.js (nyc)][nyc] has been added for code coverage. On each `yarn test`, nyc will output a summary in the terminal. In addition, a more detailed HTML version will output to the `/coverage` directory. Simply run `/coverage/index.html` in your browser to view the HTML version.
```
arielmelendez, line 63: nit: `for code coverage reporting`.

README.md
```diff
@@ -41,6 +42,157 @@ yarn install --check-cache
...
+
+Alternatively, you can view a verbose coverage output in the terminal by running:
+
+```shell
+yarn coverage
+```
+
+### Adding tests
+
+There are many different syntax options available with the Chai library, which can be found in their [documentation][chai-doc]. Here is an example unit test for a basic input/output function using Mocha and Chai:
```
arielmelendez, line 73: Rather than putting code examples here, it might be best to put them in actual source files in the correct folders. To that point, let's be instructive here about where test files should live. I believe the pattern we're looking to use is to colocate unit test files next to their analogous source files and to put integration (and maybe end-to-end?) test files in analogous folder structures under the ./tests folder.

arielmelendez: When we create the example test files, it might be good for us to advise best practices around test setup and naming conventions (e.g. I like that you've used terms like input, expectedOutput, etc.)

README.md
```diff
@@ -41,6 +42,157 @@ yarn install --check-cache
...
+        return expect(responseObject).to.deep.equal(expectedOutput);
+    });
+});
+```
+
+### Using Sinon
+
+Sinon can be used to create spies, mocks, fakes, stubs, and more. There are only some basic examples shown below, visit their [documentation][sinon-doc] to learn more.
+
+To leverage the Sinon library, simply import the methods you need into your test file:
```
arielmelendez, line 102: All of this could live in the example test files.

README.md
```diff
@@ -41,6 +42,157 @@ yarn install --check-cache
...
+
+[Power-Assert] is setup as another testing tool. The library can be used to provide a very detailed output of your failing test cases. This can become super useful while debugging a test.
+
+To use this tool, it must be imported using this syntax:
+
+```ts
+import assert = require('assert');
+```
+
+Then use `assert` in your error throwing test case:
```
arielmelendez, line 165: May want to note that you'd be REPLACING (commenting out) your existing failing assert and replacing with power assert syntax.

README.md
```diff
@@ -41,6 +42,157 @@ yarn install --check-cache
...
+
+And finally, to view the detailed error output in your terminal:
+
+```shell
+yarn power-assert -g 'My test case'
+```
+
+Here is some more [information on power-assert's options][power-assert-api].
+
+[mocha]: https://github.com/mochajs/mocha
```
arielmelendez, line 186: Consider a sub-heading for these links

README.md
arielmelendez: We might also want to point out where the configuration files live.

arielmelendez (changes requested):
See my comments. Nice work so far! Thanks.

README.md
fedellen: The links are working on my end. This is a little markdown trick, they attach to the link variables listed below.

This is just a style habit I've picked up for re-usability, I can move the links into the content.

README.md
fedellen: These are hidden when viewing the produced markdown file. They are the link variables I mentioned above, only viewable in the raw file. I will relocate them all into the content

README.md
arielmelendez: > This is just a style habit I've picked up for re-usability, I can move the links into the content.

No need. I just wasn't aware of that markdown feature. Thanks for the heads up!

README.md
fedellen: I do have an `example.test.ts` file set up in the /test directory (I can change that to /tests). These examples are mostly based on the code in that file. That example was meant to be instructive, to test that the libraries were functional with our codebase, and to make sure our test command would reach that directory.

I will be expanding that example file to include the information I added to the README. And I'll clean up that file to use some better practices. Its actually a unit test that has been split up into various examples, and doesn't really belong in the integration test directory. I just needed some test in that directory to prove it was working.

I'll also add a section in the README on where each type of tests should live. I think everyone is in agreement with you on where to locate tests.

README.md
agsuy: I think it would be easier to edit the file in the future if we keep all the links at the end rather than with the traditional [name](link) format. We just need to make sure we do that for all the links tho.

README.md
```diff
@@ -18,8 +18,97 @@ We use nvm to manage our Node engine version and, if necessary, to install an np
...
+├── src
+│   ├── fileToTest.ts
+│   └── fileToTest.test.ts   <-- Unit test
+└── tests
+    └── bestApi.test.ts   <----- Integration test
+```
+
+### Using Sinon
+
+Sinon can be used to create spies, mocks, fakes, stubs, and more. There are some basic examples of using the library shown files shared above.
```
arielmelendez, line 85: Nit: "shown files" -> "in the shown files" or similar

README.md
```diff
@@ -18,8 +18,97 @@ We use nvm to manage our Node engine version and, if necessary, to install an np
 </ol>
 
 ## Building the Library
+
 Simply run `yarn build`. This will clean the project and compile the TypeScript library.
 
+## Testing the Library
```
arielmelendez, line 24: This whole section is amazing.

agsuy:
You might want to add to yarn clean any cache/files generated by test suites.

arielmelendez (approved):

PR #59, opened 2021-07-09
PE-50: Fix Duplicate ArDrive Fee

### LINKED ISSUE #57: Duplicate fee payment
It seems that the logic for uploading files pays fees twice:  [Once for each file uploaded](https://github.com/ardriveapp/ardrive-core-js/blob/11080af8ae5748d7a7ef6283a9de7cda433a2fbc/src/public/arfs.ts#L402) and [once at the end of a batch](https://github.com/ardriveapp/ardrive-core-js/blob/11080af8ae5748d7a7ef6283a9de7cda433a2fbc/src/public/arweave.ts#L468).

The first of these is particularly annoying, since it is quite slow to retrieve the current fee rate from the smart contract and determine the token holder to pay.  In effect, this can then be the largest bottleneck (far more so than the actual upload) for uploading a lot of smallish files.

So most likely, the first call to `sendArDriveFee` should just be removed, as the total fee is paid at the end anyway?

fedellen:
This PR fixes #57  on the `dev` branch

`uploadArFSFileData` does not need to handle the fee in this case, because `uploadArDriveFiles` handles fees for all files at the end of an upload batch. Handling the fee per individual file is unnecessary here.

I self tested this with some awkward results, but with this hotfix the cost is always LESS than expected now rather than MORE. All .png uploads tested were around 300 bytes. The Cost was how much AR was taken from the active wallet when the transactions had been settled. The expected result was taken from the CLI's estimation upon confirming (y/n):

## Before hotfix 

- heart.png
  - Cost: 0.000026889017 AR 
  - Expected: 0.000012319000000000002 
  - Difference: 0.000014570017 AR **MORE** than expected
  
- coin.png
  - Cost: 0.000026825254 AR
  - Expected: 0.000012255 AR
  - Difference: 0.000014570254 AR **MORE** than expected

## After hotfix 

- key.png
  - Cost: 0.000003503865 AR 
  - Expected: 0.000012263000000000001 
  - Difference: 0.000008759135 AR **LESS** than expected
  
- barrel.png
  - Cost: 0.000003566591AR
  - Expected: 0.000012324000000000001 AR
  - Difference: 0.000008757409 AR **LESS** than expected


agsuy:
How are you calculating the expected fee?

fedellen:
> How are you calculating the expected fee?

Its the estimation from the cli as you confirm upload. I'm going to be digging into where and how that is calculated. But the AR is deducted when the transaction is settled, so there is a natural delay between estimation and payment.

agsuy:
> > How are you calculating the expected fee?
> 
> Its the estimation from the cli as you confirm upload. I'm going to be digging into where and how that is calculated. But the AR is deducted when the transaction is settled, so there is a natural delay between estimation and payment.

1. So is this a defect on the way we calculate fees at CLI level that we want to report as a defect?
2. Its worth using CLI calculations as reference if we know they are wrong?

fedellen:
> 1. So is this a defect on the way we calculate fees at CLI level that we want to report as a defect?
> 2. Its worth using CLI calculations as reference if we know they are wrong?

1. After digging further I learned that there is another defect in the way we are estimating fees here. The method is calculating 1003210 bytes in a MB, when there are actually 1048576 bytes in a MB:

```ts
// From function `getPriceOfNextUploadBatch` in `node.ts` 

// Get all files that are ready to be uploaded
const filesToUpload: types.ArFSFileMetaData[] = getDb.getFilesToUploadFromSyncTable(login);
if (Object.keys(filesToUpload).length > 0) {
  // Estimate the size by getting the size of 1MB
  const priceFor1MB = await getWinston(1000000);
  const pricePerByte = priceFor1MB / 1003210; // <-- There are actually 1048576 bytes in a MB
```

The method also makes an assumption for the metadata price (in multiple places) which could be effecting it's accuracy:

```ts
totalArweaveMetadataPrice += 0.0000005; // Ths is the price we assume it costs for a metadata tx
```

2. I believe if we fix the defect above the estimation will be more accurate. However, the estimation will always have this natural delay and should still be slightly inaccurate, especially when uploading a LOT of large files. 

Having an estimation is way more user friendly than not having an estimation, but I think we should at least warn the user that its just an estimation and AR price is volatile and "could" shift within the 2-15 minutes that an upload can take to settle.

Also, I found that during an upload there is an additional estimation that is shown to the user in method `uploadArFSFileData` of `arfs.ts`. I haven't been using this one in my expected results because there is no prompt to confirm the upload at the time it is shown. But, this estimation does not include the ArDrive community fee -- so it is underestimating the cost as well.

I'll ready a couple commits to clear things up here

arielmelendez:
> This PR fixes #57 on the `dev` branch
> 
> `uploadArFSFileData` does not need to handle the fee in this case, because `uploadArDriveFiles` handles fees for all files at the end of an upload batch. Handling the fee per individual file is unnecessary here.
> 
> I self tested this with some awkward results, but with this hotfix the cost is always LESS than expected now rather than MORE. All .png uploads tested were around 300 bytes. The Cost was how much AR was taken from the active wallet when the transactions had been settled. The expected result was taken from the CLI's estimation upon confirming (y/n):
> 
> ## Before hotfix
> * heart.png
>   
>   * Cost: 0.000026889017 AR
>   * Expected: 0.000012319000000000002
>   * Difference: 0.000014570017 AR **MORE** than expected
> * coin.png
>   
>   * Cost: 0.000026825254 AR
>   * Expected: 0.000012255 AR
>   * Difference: 0.000014570254 AR **MORE** than expected
> 
> ## After hotfix
> * key.png
>   
>   * Cost: 0.000003503865 AR
>   * Expected: 0.000012263000000000001
>   * Difference: 0.000008759135 AR **LESS** than expected
> * barrel.png
>   
>   * Cost: 0.000003566591AR
>   * Expected: 0.000012324000000000001 AR
>   * Difference: 0.000008757409 AR **LESS** than expected

When expressing really small AR values, it can be helpful to express them in Winston with commas in the usual places to help the eye pick up magnitudes more quickly. e.g. for your first example:
heart.png
Cost: 26,889,017 Winston
Expected: 12,319,000 Winston
Difference: 14,570,017 Winston **MORE** than expected (~2.18x)

src/common.ts
```diff
@@ -514,3 +514,7 @@ export async function getArUSDPrice(): Promise<number> {
 		return 0;
 	}
 }
+
+export function winstonToAr(winston: number): number {
```
arielmelendez, line 518: We should throw an exception if the absolute value of the Winston number is less than 1.0 since there's no such thing as fractional Winston; it's the smallest unit of AR.

agsuy: For reference: https://docs.arweave.org/developers/server/http-api#ar-and-winston

src/node.ts
```diff
@@ -197,37 +203,42 @@ export async function getPriceOfNextUploadBatch(login: string): Promise<types.Up
...
 			return 'Calculated price';
 		});
 
-		// Calculate the total price for all files/folders
-		const totalArweaveDataPrice = totalWinstonData * 0.000000000001;
+		// Convert the total winston data into total AR price for all files/folders
+		const totalArweaveDataPrice = common.winstonToAr(totalWinstonData);
 
-		// Add the ArDrive fee
+		// Add the ArDrive community fee
```
arielmelendez, line 238: 👍

src/constants.ts
```diff
@@ -11,5 +11,7 @@ export const appVersion = '0.1.0';
 export const arFSVersion = '0.11';
 export const defaultCipher: CipherType = 'AES256-GCM';
 
+export const minimumArDriveARFee = 0.00001; // AR
```
arielmelendez, line 14: Could be nice to include a link here to the provenance of this value (i.e. the Community Improvement Proposal Doc https://arweave.net/Yop13NrLwqlm36P_FDCdMaTBwSlj0sdNGAC4FqfRUgo that was voted upon recently  - vote #82 at https://community.xyz/#-8A6RexFkpfWwuyVO98wzSFZh0d6VJuI-buTJvlwOJQ/votes/)

arielmelendez: Nit: "Tip" is the common word that's used.

src/smartweave.ts
```diff
@@ -14,7 +14,7 @@ export async function getArDriveFee(): Promise<number> {
 		);
 		return arDriveCommunityFee ? arDriveCommunityFee[1] : 15;
 	} catch {
-		return 0.15; // Default fee of 15% if we cannot pull it from the community contract
+		return 15; // Default fee of 15% if we cannot pull it from the community contract
```
arielmelendez, line 17: This is another magic number worth factoring out as it's used here and on line 15 above. Additionally, the function name should be updated to include the UNIT of the returned value. In this case it's a percentage. And this function's value is that it is getting the fee from the contract, so we may further want to include that so that we can disambiguate it from a function that returns the fee in Winston when provided a data size in bytes.

src/public/arfs.ts
```diff
@@ -325,6 +327,9 @@ export async function uploadArFSFileData(
 	try {
 		const winston = await getWinston(fileToUpload.fileSize);
 		arPrice = +winston * 0.000000000001;
```
arielmelendez, line 329: Let's factor that multiplier out and have a constant like `arPerWinston`.

src/public/arfs.ts
```diff
@@ -325,6 +327,9 @@ export async function uploadArFSFileData(
...
+		// Add ArDrive Fee estimation
+		const arDriveFee = arPrice * ((await getArDriveFee()) / 100);
```
arielmelendez, line 331: It might be best to return a decimal fraction value from whatever getArDriveFee is going to be renamed to in order to prevent callers from having to remember to divide by 100.

src/node.ts
```diff
@@ -197,37 +203,42 @@ export async function getPriceOfNextUploadBatch(login: string): Promise<types.Up
...
 		let arDriveFee = +totalArweaveDataPrice.toFixed(9) * ((await getArDriveFee()) / 100);
```
arielmelendez, line 239: We get 12 digits of AR precision when working with Winstons. Why is this `toFixed(9)`?

src/common.test.ts
```diff
@@ -70,3 +70,13 @@ describe('The formatBytes function ', () => {
 		expect(formatBytes(1073741823)).to.equal('1024.000 MB');
 	});
 });
+
+/** Winston to AR multiplies by 0.000000000001*/
+describe('The winstonToAr function', () => {
```
arielmelendez, line 75: Test cases to consider:
1. Negative values (OK)
2. ANY non-integer value of Winston is an exception
3. 0 (zero is valid)
4. 1 AR = 1,000,000,000,000 Winston
5. 0.999999999999 = 999,999,999,999 Winston

arielmelendez (changes requested):
See my comments, requests, and questions. Thanks!

src/common.ts
fedellen: Is it worth it here to throw an exception for this case or should we simply round the number to the nearest integer before attempting conversion? 

Having it round for us would give the function a little more utility and become easier to use outside of a try/catch block. Ideally, the Arweave gateway will only ever return us a whole winston number. So the only way we normally would end up with a decimal is from some type of conversion we made in the code after we have obtained that value from the gateway.

A decimal of a winston is such a small fraction I'm just not sure its "worth" throwing errors.

So @arielmelendez , do you think this function should be:

```ts
// With exception
export function winstonToAr(winston: number): number {
  if (Number.isInteger(winston)) throw new Error(`Winston value not an integer: ${winston}`);
  return winston * 0.000000000001;
}
```

Or:

```ts
// With rounding
export function winstonToAr(winston: number): number {
  return Math.round(winston) * 0.000000000001;
}
```

src/node.ts
fedellen: Thats a good question. I'll ask Phil if there were any technical limitations behind the decision of using 9 digits

src/common.test.ts
```diff
@@ -71,12 +71,37 @@ describe('The formatBytes function ', () => {
 	});
 });
 
-/** Winston to AR multiplies by 0.000000000001*/
+/**
```
arielmelendez, line 74: This docblock seems redundant and brittle (tests might be added/changed/removed and you'd have to come update again here) since the tests are self describing.

src/common.test.ts
```diff
@@ -71,12 +71,37 @@ describe('The formatBytes function ', () => {
...
+ * Correctly represents 1 AR
+ * Correctly represents 0.999999999999 AR
+ * Removes trailing 0s
+ * Works with a Winston value of 0
+ * Throws error when Winston value is a decimal
+ */
 describe('The winstonToAr function', () => {
 	it('correctly converts winston data price to AR token price', () => {
-		expect(winstonToAr(62345548231)).to.equal(0.062345548231);
+		expect(winstonToAr(62_345_548_231)).to.equal(0.062_345_548_231);
```
arielmelendez, line 89: Ooooh numeric separators. Sexy. Can't say I've ever seen that in the wild, but I'm into it :D

src/common.ts
arielmelendez: Yea we'll see how this goes. Thanks for the update.

src/public/arweave.ts
```diff
@@ -54,23 +54,23 @@ export async function uploadDataChunk(uploader: TransactionUploader): Promise<Tr
...
 	try {
 		// Get the latest ArDrive Community Fee from the Community Smart Contract
-		let fee = arPrice * ((await getArDriveFee()) / 100);
+		let tip = arPrice * (await getArDriveTipPercentage());
 
-		// If the fee is too small, we assign a minimum
-		if (fee < 0.00001) {
-			fee = 0.00001;
+		// If the tip is too small, we assign a minimum
+		if (tip < minimumArDriveCommunityTip) {
```
arielmelendez, line 64:
```
tip = Math.max(tip, minimumArDriveCommunityTip);
```
might be a bit more concise.

src/public/arfs.ts
arielmelendez: Bump

arielmelendez (changes requested):
Just one more constant to factor out. And please follow up on the toFixed(9) issue.

.mocharc.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
 	extension: ['ts'],
 	spec: ['**/*.test.ts'],
 	require: ['ts-node/register/transpile-only', 'source-map-support/register', 'tests/testSetup.ts'],
-	timeout: '3000',
+	timeout: '10000', // 10 seconds for smartweave contract reading
```
arielmelendez, line 9: Is there no way to change this value on the fly when we need to rather than extending the timeouts across the app? (I might be misinterpreting the usage of this constant)

src/common.test.ts
```diff
@@ -70,3 +70,26 @@ describe('The formatBytes function ', () => {
 		expect(formatBytes(1073741823)).to.equal('1024.000 MB');
 	});
 });
+
+// winstonToAr multiplies by 0.000000000001
```
arielmelendez, line 74: I think even this small comment is redundant 😬

src/common.ts
```diff
@@ -514,3 +514,7 @@ export async function getArUSDPrice(): Promise<number> {
 		return 0;
 	}
 }
+export function winstonToAr(winston: number): number {
+	if (!Number.isInteger(winston)) throw new Error(`Winston value not an integer: ${winston}`);
```
arielmelendez, line 518: A thought.... isInteger gives us 64 bits of precision to work with. E.g.
```
Number.isInteger(5.000000000000001); // false
Number.isInteger(5.0000000000000001); // true
```
All that means is that tiny fractional amounts of Winston might get treated as integers but ALSO large amounts of Winston with some large fractional amount might also be considered an integer. That's probably ok.

fedellen: Thats interesting. Do you know any good workaround for this?

I just tried using `winston % 1 !== 0` and also `winston.toString().indexOf('.') !== -1` but both are similarly failing for the above test case.

JavaScript seems to be saying exactly `5.0000000000000001 === 5`. And it seems to also round up if that last digit is meant to: `5.000_000_000_000_000_5 === 5.000_000_000_000_001`

.mocharc.js
fedellen: I did find a way to change it on the fly and updated it.

We can directly change settings for a mocha test in the describe block:

```ts
describe('The estimateArCost function', function () {
  // Set timeout to 10 seconds for smartweave contract reading
  this.timeout(10000);
  
  // ...

```

fedellen:
This all should have been a different branch with a new JIRA ticket (I'll watch my scope creep next time, just started digging around) -- but in this branch I recently made two significant changes to the community tip fetching and the estimation process.

## Community Tip

`getArDriveFee` is now `getArDriveTipPercentage` and will grab tip percentage from the community vote at it's block height of `578358`. This alone reduces the time to read the contract from 40000ms-60000ms to 4000ms-6000ms. In addition it will save this value for that session, causing synchronous retrieval on subsequent calls.

If the fee has not been fetched from the exact setting (the 40-60 second contract read), it will also start a background task to go fetch that setting. This task will update the value retrieved above when it finishes. 

The way I set up this background task is the first time I've tried "not awaiting" code like this. I've done a good amount of self testing with the Core -> CLI pipeline and I'm not running into any problems as of yet. But I'm not overly confident about this background task because I haven't used this pattern before. The [code is located](https://github.com/ardriveapp/ardrive-core-js/pull/59/files#diff-7015ad9d4d41d585b726513fb0bfbe45a9916e1c18c321721daf6e2aecd5aabe) in `smartweave.ts` 

## Estimation

The estimation process has been heavily refactored and slightly adjusted. I factored out two functions that have been shared into other places of the code: `estimateArCost` and `getArDriveCommunityTip`. In doing this I simplified and documented some of the logic inside the main price estimation function: `getPriceOfNextUploadBatch`.

One change to note here is that I have increased the `assumedMetadataTxPrice` from `0.000_000_500_000` to `0.000_002_500_000`.  I found that the meta data estimations have been far too low while working with tiny files.

### Latest estimations in Winston

Overall the estimations have become a lot more accurate, and generally estimate slightly more than the actual value taken.

- lilesplode.png (263 bytes)
  - estim : 15,363,052 
  - actual: 15,093,884
  - diff : 269,168 Winston **LESS** expected
 
- larger_file.png (53.908 KB)
  - estim : 44,149,059 
  - actual: 43,379,377
  - diff : 769,682 Winston **LESS** expected     
 
- ui_heart.png + ui_key.png (2 files, 724 bytes)
  - estim : 20,317,814 
  - actual: 18,772,287
  - diff : 1,545,527 Winston **LESS** expected
  
- bat_animation.png (262 bytes)
  - estim : 15,931,875 
  - actual: 15,162,193
  - diff : 769,682 Winston **LESS** expected

src/common.ts
arielmelendez: It's just a limitation of the number system. For 0-9AR (0-~9.9 trillion Winston), we'll get about 4-5 digits of precision here. That's probably fine.

src/constants.ts
```diff
@@ -11,5 +11,17 @@ export const appVersion = '0.1.0';
...
 export const defaultCipher: CipherType = 'AES256-GCM';
 
+/**
+ * Minimum ArDrive community tip from the Community Improvement Proposal Doc:
+ * https://arweave.net/Yop13NrLwqlm36P_FDCdMaTBwSlj0sdNGAC4FqfRUgo
+ *
+ * Voted on by the ArDrive community (vote #82):
+ * https://community.xyz/#-8A6RexFkpfWwuyVO98wzSFZh0d6VJuI-buTJvlwOJQ/votes/
+ */
+export const minimumArDriveCommunityTip = 0.000_010_000_000;
```
arielmelendez, line 21: Nit: Knowing what unit it's in from the could be helpful. e.g. `minArDriveCommunityTipAR`

src/constants.ts
```diff
@@ -11,5 +11,17 @@ export const appVersion = '0.1.0';
...
+
+/** Estimated AR price for most metadata transactions */
+export const assumedMetadataTxPrice = 0.000_002_500_000;
```
arielmelendez, line 24: Same idea here, e.g. `assumedMetadataTxARPrice`

src/node.test.ts
```diff
@@ -0,0 +1,15 @@
+import { expect } from 'chai';
+import { estimateArCost } from './node';
+
+describe('The estimateArCost function', function () {
+	// Set timeout to 10 seconds for smartweave contract reading
+	this.timeout(10000);
```
arielmelendez, line 6: We shouldn't write unit tests that hit the network. Only end-to-end tests should do that. We can redesign this function for testability if we provide for an object or function type that retrieves pricing data as a parameter to estimateArCost (with a sensible default parameter value for most normal use cases). If we do that, then we can stub the response from it and make sure the returned value is exactly what we expect.

src/node.ts
```diff
@@ -185,57 +185,67 @@ export async function getPriceOfNextUploadBatch(login: string): Promise<types.Up
...
 }
+
+export async function estimateArCost(totalSize: number, numberOfFiles = 1): Promise<number> {
+	// Extra bytes added to the header of data uploads
+	const headerByteSize = 3210;
+	const sizeWithHeaders = totalSize + numberOfFiles * headerByteSize;
+	// Get Winston value from gateway, and convert to AR for all files/folders to be uploaded
+	const arCost = common.winstonToAr(await getWinston(sizeWithHeaders));
+	// Return cost, with added community tip
+	return arCost + (await getArDriveCommunityTip(arCost));
```
arielmelendez, line 244: `getArDriveCommunityTip` is the piece that needs to be factored out/parameterized so that you can unit test this function. You need an object/interface type like `ArCommunityTipReader` which you could then mock/stub in your unit test.

src/smartweave.ts
```diff
@@ -4,17 +4,57 @@ import { arweave } from './public/arweave';
...
-export async function getArDriveFee(): Promise<number> {
+// Default tip of 15% if we cannot pull it from the community contract
+const defaultArDriveTipPercentage = 0.15;
+let arDriveTipPercentage = 0;
+let isTipPercentageFromSettings = false;
+
+// Calls the ArDrive Community Smart Contract to pull the tip
+export async function getArDriveTipPercentage(): Promise<number> {
+	// If tip was not gathered from exact setting, send background task to gather, do not await
+	if (!isTipPercentageFromSettings) getExactArDriveTipSetting();
```
arielmelendez, line 17: Style nit: Move the code execution into braces and on its own line.

src/smartweave.ts
```diff
@@ -4,17 +4,57 @@ import { arweave } from './public/arweave';
...
 		const arDriveCommunityFee = contract.settings.find(
 			(setting: (string | number)[]) => setting[0].toString().toLowerCase() === 'fee'
 		);
-		return arDriveCommunityFee ? arDriveCommunityFee[1] : 15;
-	} catch {
-		return 0.15; // Default fee of 15% if we cannot pull it from the community contract
+
+		arDriveTipPercentage = arDriveCommunityFee;
+		isTipPercentageFromSettings = true;
+		isTipFromSettingsLoading = false;
```
arielmelendez, line 54: Consider using a `finally` block to set `isTipFromSettingsLoading` to `false`.

arielmelendez (changes requested):
A few minor things. But the major one is not to hit the network during unit tests.

src/common.test.ts
```diff
@@ -70,3 +70,26 @@ describe('The formatBytes function ', () => {
...
+describe('The winstonToAr function', () => {
+	it('correctly converts winston data price to AR token price', () => {
+		expect(winstonToAr(62_345_548_231)).to.equal(0.062_345_548_231);
+	});
+	it('correctly coverts with negative winston values', () => {
```
arielmelendez, line 78: Nit: coverts -> converts

src/node.test.ts
```diff
@@ -0,0 +1,34 @@
...
+	arweaveOracleStub.getWinstonPriceForByteCount.returns(
+		new Promise((resolve) => {
+			resolve(12_345_678);
+		})
+	);
+
+	const communityOracleStub = stubInterface<CommunityOracle>();
+	communityOracleStub.getCommunityARTip.returns(
+		new Promise((resolve) => {
+			resolve(minArDriveCommunityARTip);
```
arielmelendez, line 19: So if the minimum AR tip amount changes, the tests below will fail. That could be good or bad depending on what you're trying to achieve. Since the focus of the test ins't making sure that the minimum value is something specific, I'd recommend using a numeric literal here rather than importing the constant.

src/node.ts
```diff
@@ -185,57 +188,69 @@ export async function getPriceOfNextUploadBatch(login: string): Promise<types.Up
...
 			if (!common.checkFileExistsSync(fileToUpload.filePath)) {
 				console.log('%s is not local anymore.  Removing from the queue.', fileToUpload.filePath);
 				await deleteFromSyncTable(fileToUpload.id);
 				return 'File not local anymore';
 			}
+
 			// Calculate folders that are ready to be uploaded, but have no TX already
 			if (+fileToUpload.fileMetaDataSyncStatus === 1 && fileToUpload.entityType === 'folder') {
-				totalArweaveMetadataPrice += 0.0000005; // Ths is the price we assume it costs for a metadata tx
+				totalArweaveMetadataPrice += assumedMetadataTxARPrice;
```
arielmelendez, line 202: It still surprises me that we can't precompute the total metadata size, divide by the number of files, get an AR quote on that, and then multiply by the number of files again to get a pretty close estimate.

src/public/ardrive_community_oracle.ts
```diff
@@ -0,0 +1,89 @@
...
+import { CommunityOracle } from './community_oracle';
+
+// Default tip of 15% if we cannot pull it from the community contract
+const defaultArDriveTipPercentage = 0.15;
+let arDriveTipPercentage: number | undefined = undefined;
+
+let isTipPercentageLatestFromSmartWeave = false;
+let isFetchingTipSetting = false;
+
+const communityTipBlockHeight = 578358;
```
arielmelendez, line 14: This is worth a comment explaining the value.

src/public/ardrive_community_oracle.ts
```diff
@@ -0,0 +1,89 @@
...
+
+	// Calls the ArDrive Community Smart Contract to pull the tip percentage
+	async getArDriveTipPercentage(): Promise<number> {
+		if (!isTipPercentageLatestFromSmartWeave) {
+			// If tip was not gathered from exact setting, start background task to gather, do not await
+			this.setExactTipSettingInBackground();
+		}
+
+		try {
+			if (arDriveTipPercentage !== undefined) {
```
arielmelendez, line 30: Moved this outside the try.

src/public/ardrive_community_oracle.ts
```diff
@@ -0,0 +1,89 @@
...
+	/**
+	 * Derives the exact setting from smartweave contract (35-50+ seconds)
+	 * This function is to be used in the background and not awaited on
+	 * When the setting is derived, it will set it in the cache (arDriveTipPercentage)
+	 *
+	 * @TODO Ideally this task should begin on start up to have the best
+	 *       chance of retrieving the exact value before the user needs it
+	 */
+	async setExactTipSettingInBackground(): Promise<void> {
+		while (isFetchingTipSetting) {
```
arielmelendez, line 61: I think we can remove this loop now since only 1 caller at a time will ever actually try to get the data.

arielmelendez: This was only ever necessary if we expected callers to await on it, but your doc block says NOT to use it that way.

src/public/gateway_oracle.ts
```diff
@@ -0,0 +1,8 @@
+import { getWinston } from '../node';
+import { ArweaveOracle } from './arweave_oracle';
+
+export class GatewayOracle implements ArweaveOracle {
+	getWinstonPriceForByteCount(byteCount: number): Promise<number> {
+		return getWinston(byteCount);
```
arielmelendez, line 6: Consider moving the implementation of getWinston here and replacing calls to it in bundles.ts with calls to a GatewayOracle object.

arielmelendez (changes requested):
Major issue: Why does using `votes` work?
Minor issues: A few in my other comments.

fedellen:
> Major issue: Why does using `votes` work?

Because using the block height version of the contract read we are searching through the `votes` array for the last vote. The setting for fee hasn't been created on that version of the contract. This is what the contract looks like during this block height contract read (minimized):

```
{
  name: 'ArDrive',
  ticker: 'ARDRIVE',
  balances: {
     // ...
   },
   vault: {
     // ...
   },
   votes: [
      // ...
      },
      // This is the last vote when using block height 578358
      {
        status: 'active',
        type: 'set',
        note: 'Setting ArDrive Data Upload fee',
        yays: 0,
        nays: 0,
        voted: [],
        start: 578358,
        totalWeight: 1603684484400,
        key: 'fee',
        value: 15
      }
   ],
   roles: {},
   settings: [
      [ 'quorum', 0.5 ],
      [ 'support', 0.15 ],
      [ 'voteLength', 2160 ],
      [ 'lockMinLength', 720 ],
      [ 'lockMaxLength', 788400 ],
      [ 'communityAppUrl', 'https://ardrive.io' ],
      [ 'communityDiscussionLinks', [Array] ],
      [
        'communityDescription',
        "We are a community focused on building the best private, secure, decentralized, pay-as-you-go, censorship-resistant and permanent data storage solution, for everyone.  With ArDrive's desktop, mobile and web apps, you can easily sync and share your public and private files from the PermaWeb."
      ],
      [ 'communityLogo', '1xQvwhgFF7573l5Ejb-5DHudo16R7aF0cTEKJMwsLAs' ]
   ]
}
```

src/node.ts
fedellen: I've been focused on the estimation/testing side of this and haven't fully investigated the Metadata process. I can still dig further into this one and see if I can find a way, or at least find the reason why we can't

src/public/ardrive_community_oracle.ts
```diff
@@ -58,26 +60,23 @@ export class ArDriveCommunityOracle implements CommunityOracle {
...
 	 */
 	async setExactTipSettingInBackground(): Promise<void> {
-		while (isFetchingTipSetting) {
-			// Do not run if contract is currently being read
-			await sleep(500);
-		}
-
-		if (isTipPercentageLatestFromSmartWeave) {
-			// Exact tip percentage has already been derived, return early
+		if (isTipPercentageLatestFromSmartWeave || isFetchingTipSetting) {
```
arielmelendez, line 63: Good catch

src/public/ardrive_community_oracle.ts
```diff
@@ -58,26 +60,23 @@ export class ArDriveCommunityOracle implements CommunityOracle {
...
+			// the contract is currently being read, return early
 			return;
 		}
 
 		isFetchingTipSetting = true;
 
 		try {
 			const contract = await readContract(arweave, communityTxId);
-			const arDriveCommTipFromSettings: number = contract.settings.find(
+			const arDriveCommTipFromSettings: ['fee', number] = contract.settings.find(
```
arielmelendez, line 73: This is the kind of change we would have realized we needed earlier if we factored out the contract reader and unit tested this code. We would have been forced to learn the schema of the returned data and tested against it.

arielmelendez (approved):
Good improvements. Thanks!

jdaev (dismissed):
Looks good to me

.mocharc.js
```diff
@@ -1,5 +1,7 @@
 'use-strict';
 
+process.env.NODE_ENV = 'test';
```
arielmelendez, line 3: Is this still required now that you removed the conditional code that relied upon this value?

fedellen: No, but it might be useful addition to the test setup

src/public/gateway_oracle.ts
```diff
@@ -0,0 +1,9 @@
+import { ArweaveOracle } from './arweave_oracle';
+import fetch from 'node-fetch';
+
+export class GatewayOracle implements ArweaveOracle {
+	async getWinstonPriceForByteCount(byteCount: number): Promise<number> {
+		const response = await fetch(`https://arweave.net/price/${byteCount}`);
+		return response.json();
```
arielmelendez, line 7: Let's make sure this is returning the correct type.

src/bundles.ts
```diff
@@ -60,9 +60,11 @@ export async function uploadArDriveFilesAndBundles(user: types.ArDriveUser): Pro
 					const fileDataItem: DataItemJson | null = await createArFSFileDataItem(user, filesToUpload[n]);
 					if (fileDataItem !== null) {
 						// Get the price of this upload
-						const winston = await getWinston(filesToUpload[n].fileSize);
+						const winston = await new GatewayOracle().getWinstonPriceForByteCount(
+							filesToUpload[n].fileSize
+						);
 						totalSize += filesToUpload[n].fileSize;
-						totalARPrice += +winston * 0.000000000001; // Sum up all of the fees paid
+						totalARPrice += common.winstonToAr(+winston); // Sum up all of the fees paid
```
arielmelendez, line 67: Once we verify that getWinstonPriceForByteCount is returning the correct type, let's remove the `+` in `+winston`.

src/public/ardrive_community_oracle.test.ts
```diff
@@ -0,0 +1,55 @@
+import { expect } from 'chai';
+import { stub } from 'sinon';
+import { stubInterface } from 'ts-sinon';
+import { minArDriveCommunityARTip } from '../constants';
+import { ArDriveCommunityOracle, communityTipBlockHeight } from './ardrive_community_oracle';
+import { ContractOracle } from './contract_oracle';
+
+describe('The ArDriveCommunityOracle class', () => {
+	it('returns a decimal to be used as a percentage using the `getArDriveTipPercentage` function', async () => {
```
arielmelendez, line 9: Nit: I think the test would read ok with a description of:
'`getArDriveTipPercentage` function returns a decimal to be used as a percentage'

arielmelendez: The same applies to the tests below.

src/public/ardrive_community_oracle.ts
```diff
@@ -0,0 +1,76 @@
...
+			console.log(err);
+			return defaultArDriveTipPercentage;
+		}
+	}
+
+	/**
+	 * Fetches and caches the exact setting from SmartWeave contract (35-50+ seconds)
+	 * This function is to be used in the background and not awaited on
+	 */
+	async setExactTipSettingInBackground(
```
arielmelendez, line 61: Any thoughts on whether this function should run if the cachedArDriveTipPercentage is set already?

src/public/ardrive_community_oracle.test.ts
```diff
@@ -0,0 +1,55 @@
...
+
+		const output = await new ArDriveCommunityOracle().setExactTipSettingInBackground(smartWeaveOracleStub);
+
+		expect(output).to.equal(5.67);
+		expect(smartWeaveOracleStub.getCommunityTipSetting.calledOnceWithExactly()).to.be.true;
+
+		return;
+	});
+
+	it('returns the `minArDriveCommunityARTip` if the calculated community tip is a smaller value', async () => {
```
arielmelendez, line 41: There should be a complimentary test that if the tip is larger than the minimum that it returns that value. Also the function name should be in the test description. Would it help if you used a nested describe block for each batch of  test cases for each function?

src/public/ardrive_community_oracle.ts
```diff
@@ -0,0 +1,76 @@
...
+		if (cachedArDriveTipPercentage !== undefined) {
+			// Tip percentage has already been fetched, return that value
+			return cachedArDriveTipPercentage;
+		}
+
+		try {
+			// Tip has not been calculated, use block height to read from contract (4-6 seconds)
+			const communityTipValue = await contractOracle.getCommunityTipSetting(communityTipBlockHeight);
+
+			if (communityTipValue) {
```
arielmelendez, line 39: If the community tip value were voted to be 0 (oh nooooo!), this would be falsey and end up returning the hardcoded default. I'd say that you should consider checking against undefined explicitly, but the Promise is typed to deliver a number. Under the hood, you're calling through to `readContract` and I'm not sure how it behaves under adverse conditions.

src/public/smartweave_oracle.ts
```diff
@@ -0,0 +1,66 @@
+import { ContractOracle } from './contract_oracle';
+import { readContract } from 'smartweave';
+import { arweave } from './arweave';
+import { communityTxId } from '../constants';
+
+export interface CommunityContractData {
+	votes: [Record<string, unknown>];
+	settings: [string, unknown][];
+}
+export class SmartWeaveContractOracle implements ContractOracle {
```
arielmelendez, line 10: Nit: add a line of whitespace

src/public/smartweave_oracle.ts
```diff
@@ -0,0 +1,66 @@
...
+		}
+
+		return arDriveCommTipFromVotes;
+	}
+
+	/* Grabs fee directly from the settings at the bottom of the contract */
+	getTipSettingFromContractSettings(contract: CommunityContractData): number {
+		const arDriveCommTipFromSettings = contract.settings.find((setting) => setting[0] === 'fee');
+
+		if (!arDriveCommTipFromSettings) {
```
arielmelendez, line 34: Would this misbehave if the community tip were simply voted and set to 0?

arielmelendez (changes requested):
Left a few minor comments. We're near the finish line here!

src/public/ardrive_community_oracle.ts
fedellen: Thats true. This value is handled under the hood, it could in theory be a negative number as well.

Should we return the default, or error out if the value were a negative?

src/public/smartweave_oracle.ts
fedellen: No, the type for `arDriveCommTipFromSettings` is `[string, unknown] | undefined`.

The value of `arDriveCommTipFromSettings[1]` in the tuple is unknown and not being checked here

fedellen: I will add an explicit type declaration here to be more clear though

arielmelendez: AH, good call. Thanks!

fedellen: Although the function above has a similar issue where it would misbehave, I think I'll add in that the community tip cannot be a negative here

src/public/community_oracle.ts
```diff
@@ -0,0 +1,10 @@
+import { ContractOracle } from './contract_oracle';
+
+export interface CommunityOracle {
+	cachedArDriveTipPercentage: number | undefined;
```
arielmelendez, line 4: Nit: Generally I think you'd want to use getters here that you could both use internally in implementations and stub if necessary for testing.

arielmelendez (dismissed):
Great updates!

agsuy:
Public folder was mean to include everything ArFS public related.
Public drives, sharing, etc.

smartweave.ts and gateway.ts both live at /src  
All our files related to nodes and arweave interaction are placed at /src as well.

arweave.ts should be moved to /src as well. (do not recall how it ended at /src/public tbh)

Given the above I would move every file on` /src/public` to` /src` bar:

- sharing.ts
- drives.ts
- arfs.ts

agsuy (changes requested):
Please see my comments about where files should live above.
I would like to remain consistent with our current file/folder structure

agsuy (approved):
Lgtm!

arielmelendez (approved):

PR #60, opened 2021-07-09
PE-51: Change private folders to encrypt with driveKey

### LINKED ISSUE #58: Wrong encryption for folder metadata
When uploading folder metadata in a private drive, the [code currently](https://github.com/ardriveapp/ardrive-core-js/blob/11080af8ae5748d7a7ef6283a9de7cda433a2fbc/src/public/arfs.ts#L447) encrypts the data with the associated file key, while it should be the drive key.

This leads to broken folders (and orphaned files in them) both when trying to redownload the drive with the CLI and also with the web app.

fedellen:
This PR fixes #58 

Adds a check for folders in two `arfs.ts` methods, and if they are folders it will now encrypt with the driveKey rather than the invalid fileKey. Also changes `newArFSPrivateFolderMetaDataItem()` to not use derive a fileKey, as the method only handles folders.

## Before hotfix

Private folders could not be viewed in the web app. Seemingly every private drive created was immediately orphaned and unrecoverable

## After hotfix

Private folders can now be viewed and interacted with in the web app. Nested folders work, and all files are visible

### DIFF
--- a/src/public/arfs.ts
+++ b/src/public/arfs.ts
@@ -273,7 +273,7 @@ export async function createArFSFileMetaDataItem(
 		// Convert to JSON string
 		const secondaryFileMetaDataJSON = JSON.stringify(secondaryFileMetaDataTags);
 		if (fileToUpload.isPublic === 1) {
-			// Public file, do not s
+			// Public file, do not encrypt
 			dataItem = await arweave.prepareArFSMetaDataItemTransaction(user, fileToUpload, secondaryFileMetaDataJSON);
 		} else {
 			// Private file, so it must be encrypted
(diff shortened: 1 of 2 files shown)


agsuy:
Some questions.

Given that the functions involved do need some refactoring. 

1. Do we want to just change some lines in these and revisit them later on for a proper refactor? Whats the scope?
2. Do not we want to refactor/document issues (limited to the function scope) on these? 
3. Do we want to create tests for these in their current form?
4. Can we create good tests for them in their current form? e.g. functions not following SRP
5. Also, is this a Hotfix at all if we are not talking of production code?

At least leaving a TODO explaining current issues in these functions as well as possible solutions seems worthwhile as that would take advantage of the time invested in getting familiar with the code so we could perform these straight away in another iteration instead of revisiting them from zero.

arielmelendez:
> 5. Also, is this a Hotfix at all if we are not talking of production code?

Yea I'd say that until we have a release and tagging workflow for this codebase, we need not consider this a hotfix. That said, we are free to tightly scope the fix for now since this is not our highest priority initiative.

src/public/arfs.ts
```diff
@@ -443,8 +450,15 @@ export async function uploadArFSFileMetaData(user: ArDriveUser, fileToUpload: Ar
 				fileToUpload.driveId,
 				user.walletPrivateKey
 			);
-			const fileKey: Buffer = await deriveFileKey(fileToUpload.fileId, driveKey);
-			const encryptedData: ArFSEncryptedData = await fileEncrypt(fileKey, Buffer.from(secondaryFileMetaDataJSON));
+
+			// Private folders encrypt with driveKey, private files encrypt with fileKey
```
arielmelendez, line 454: Can we dry this out with a function? It's a direct repeat of the code from 285-293.

arielmelendez (changes requested):
Minor request issued above to try and DRY out the code a bit.

arielmelendez (approved):
LGTM!

PR #61, opened 2021-07-22
PE-84: Fix nyc coverage library setup

fedellen:
This PR fixes the coverage reporting output for the `nyc` package.

We can now use the latest versions of the recommended setup due to resolved issues in dependencies. The coverage output now increases as unit tests are added.


arielmelendez (approved):
:+1:

agsuy (approved):

PR #63, opened 2021-08-04
chore(version): Increment version to `0.4.0` PE-96

fedellen:
This PR increments the version to prepare for release

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -7,7 +7,7 @@ export const gatewayURL = 'https://arweave.net/';
 
 export const appName = 'ArDrive-Desktop';
 export const webAppName = 'ArDrive-Web';
-export const appVersion = '0.1.0';
+export const appVersion = '0.4.0';
 export const arFSVersion = '0.11';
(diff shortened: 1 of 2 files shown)


src/constants.ts
```diff
@@ -7,7 +7,7 @@ export const gatewayURL = 'https://arweave.net/';
 
 export const appName = 'ArDrive-Desktop';
 export const webAppName = 'ArDrive-Web';
-export const appVersion = '0.1.0';
+export const appVersion = '0.4.0';
```
arielmelendez, line 10: Looks like this got way out of date. Should we update the README with version increment instructions to try and promote awareness that both should be handled?

arielmelendez (approved):
Left a minor comment/suggestion. Approved otherwise.

src/constants.ts
fedellen: After we know right the steps, we should make a task to add a section for documenting the whole release process.

Something like:

- Increment version in `package.json` and `constants.ts`
- Merge all changes from `dev` to `staging`
- Run QA smoke testing on `staging`
- Release as NPM package

PR #67, opened 2021-08-08
Handle data cipher IV more directly

domob1812:
This streamlines the way the cipher IV for file data is handled:  Instead of querying for it during sync and with GQL, we retrieve it directly from the gateway by txid, and only during download.  This helps reduce GQL queries (and thus rate-limiting issues), and also speeds up the initial sync (especially when folders are cloud-only).

Also includes a new method for handling gateway requests / failures / retries / switching, which can be used to gradually replace the ad-hoc logic about this in the current codebase.

### DIFF
--- a/src/gateway.ts
+++ b/src/gateway.ts
@@ -1,15 +1,57 @@
 import { arweave } from './arweave';
+import Axios from 'axios';
+
+// List (in order of initial preference) of the gateways for requests.
+const gateways = [
+	"https://arweave.net",
+	"https://arweave.live",
+	"https://arweave.dev"
+];
+
+// Index of the currently used gateway into gateways.
+let currentGateway: number = 0;
+
+// Switches to the next gateway.
+function switchGateway() {
+	currentGateway = (currentGateway + 1) % gateways.length;
+	console.log("Switched gateway to " + gateways[currentGateway]);
+}
+
+// Chooses the current gateway and runs a callback with it.
+// If an error occurs, the call is retried and the gateway
+// switched automatically.
+export async function queryGateway(query: (url: string) => Promise<any>): Promise<any> {
+	const initialGatewayIndex = currentGateway;
+	let tries: number = 0;
+	while (true) {
+		try {
+			return await query(gateways[currentGateway]);
+		} catch (err) {
+			console.log(err);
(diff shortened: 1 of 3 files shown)


src/gateway.ts
```diff
@@ -1,17 +1,72 @@
...
-		return data;
-	} catch (err) {
-		console.log('Error getting transaction data for Txid %s', txid);
-		console.log(err);
-		return Promise.reject(err);
+// List (in order of initial preference) of the gateways for requests.
+const gateways = [
+	"https://arweave.net",
+	"https://arweave.live",
+	"https://arweave.dev"
```
jdaev, line 9: Is it wise to use the arweave.dev gateway? There was some discussion of inconsistent results from the arweave discord. Perhaps we could consider [arweave-multihost](https://www.npmjs.com/package/arweave-multihost)

domob1812: I don't know.  .net and .dev are what is currently used, and I was told .live is a new one.  But if you know more about this, I'm definitely happy to apply whatever changes seem best.

vilenarios: Ideally, we compare result sets from all and get some kind of baseline. We can stick with arweave.net for as the primary and only try the rest of the primary is down. 

http://arweave.live is the new Vartex gateway.  it is syncing, just like our own gateways...
http://gateway.ardrive.io
http://test-gateway.ardrive.io

domob1812: Yes, that's exactly what my code does (start with the first, and only switch to the others if some operation fails).  Do you mean the list should instead be:
- arweave.net
- arweave.live
- gateway.ardrive.io
- test-gateway.ardrive.io
(In that order?)

src/gateway.ts
```diff
@@ -1,17 +1,72 @@
...
+			    tries = 0;
+			    switchGateway ();
+			    if (currentGateway === initialGatewayIndex) {
+				    // We've tried all gateways, nothing left to do.
+				    return Promise.reject(err);
+			    }
+			}
+		}
 	}
 }
```
fedellen, line 44: These gateway fallbacks and direct axios calls are still useful and relevant to the new Core. Leaving this open for now

PR #70, opened 2021-08-11
docs: Update readme

agsuy:
Readme was outdated.

Cleaned it out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,9 +139,6 @@ yarn install --check-cache
 
 for external PR
 
-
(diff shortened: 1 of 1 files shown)


fedellen (approved):

matibat (commented):
Why is the README outdated?
We do no longer have the recommended extensions for vscode?
Maybe that doc chunk was moved elsewhere before?

fedellen:
> Maybe that doc chunk was moved elsewhere before?

Yes, it has been added above

PR #71, opened 2021-08-25
Use latest version in getFolderFromSyncTable

domob1812:
`getFolderFromSyncTable` (which is used e.g. to determine the parent folder ID for a new file) used to retrieve "any" entry with a matching path, which could be an old version of a moved folder.  In this case, files would be uploaded to a wrong location in the ArDrive.

This change makes the method return the latest matching entry by `unixTime`.  While this is not 100% perfect (it could still be wrong if there were situations where two folders had the same path and then one of them was renamed), it is a simple fix that works well most of the time.


fedellen:
Closing, feel free to move this one over to: https://github.com/ardriveapp/ardrive-sync

PR #72, opened 2021-09-15
fix(file meta data): Use application/json on content-type for meta tags PE-294

fedellen:
This PR fixes the Content-Type tag on file meta tags issue


vilenarios (approved):
Reviewed and approved!  Thanks Derek.

PR #73, opened 2021-09-15
PE-383: Update smartweave/arweave deps and version bump to 0.5.1

fedellen:
This PR bumps the versions of the smartweave and arweave dependencies and bumps core's version to 0.5.1


agsuy (approved):

src/constants.ts
```diff
@@ -7,7 +7,7 @@ export const gatewayURL = 'https://arweave.net/';
 
 export const appName = 'ArDrive-Desktop';
 export const webAppName = 'ArDrive-Web';
-export const appVersion = '0.5.0';
+export const appVersion = '0.5.1';
```
matibat, line 10: Can't we extract that value from package.json?

matibat (approved):
Looks good

src/constants.ts
fedellen: I know how to do this with `dotenv` / `dotenv-safe`:  `process.env.NPM_VERSION` `NPM_VERSION=$npm_package_version`

Just tried locally without using `dotenv` and I get undefined.

PR #79, opened 2021-11-15
PE-477: Create manifests

fedellen:
This PR adds the create manifest changes from https://github.com/ardriveapp/ardrive-cli/pull/137 to the new migrated core.


src/ardrive.ts
```diff
@@ -1002,6 +1013,133 @@ export class ArDrive extends ArDriveAnonymous {
 		};
 	}
 
+	async uploadPublicManifest({
+		folderId,
```
arielmelendez, line 1017: This API design makes it possible to pass an inconsistent folderId and driveId. Would it make more sense for us to just eat the getDriveIdForFolderId call every time to make sure we always get the right driveId?

src/ardrive.ts
```diff
@@ -1002,6 +1013,133 @@ export class ArDrive extends ArDriveAnonymous {
...
+		driveId,
+		destManifestName = 'DriveManifest.json',
+		maxDepth = Number.MAX_SAFE_INTEGER
+	}: UploadPublicManifestParams): Promise<ArFSManifestResult> {
+		if (!driveId) {
```
arielmelendez, line 1022: Everything from here to `folderId ??= drive.rootFolderId;` creates the possibility for lots of different permutations of things to happen. This function's API is borderline hard to use correctly.

arielmelendez: Consider making a private function that takes trusted driveId and folderId. Then have public functions that take one or the other but not both.

src/ardrive.ts
```diff
@@ -1002,6 +1013,133 @@ export class ArDrive extends ArDriveAnonymous {
...
+		const owner = await this.getOwnerForDriveId(driveId);
+		await this.assertOwnerAddress(owner);
+
+		const drive = await this.arFsDao.getPublicDrive(driveId, owner);
+
+		folderId ??= drive.rootFolderId;
+
+		// TODO: Handle collision with existing manifest. New manifest will always be a new file, with
+		// upsert by default this means it will only skip here on --skip conflict
+		const filesAndFolderNames = await this.arFsDao.getPublicNameConflictInfoInFolder(folderId);
```
arielmelendez, line 1039: There's a lot going on from here to the bottom of this function. Can we factor some of it out to testable helper functions and compose those together in a way that is readable and comprehensible? And then also test those?

src/ardrive.ts
```diff
@@ -1002,6 +1013,133 @@ export class ArDrive extends ArDriveAnonymous {
...
+
+		const uploadBaseCosts = await this.estimateAndAssertCostOfFileUpload(
+			arweaveManifest.size,
+			this.stubPublicFileMetadata(arweaveManifest, destManifestName),
+			'public'
+		);
+		const fileDataRewardSettings = { reward: uploadBaseCosts.fileDataBaseReward, feeMultiple: this.feeMultiple };
+		const metadataRewardSettings = { reward: uploadBaseCosts.metaDataBaseReward, feeMultiple: this.feeMultiple };
+
+		const uploadFileResult = await this.arFsDao.uploadPublicFile({
```
arielmelendez, line 1106: No dry run option?

src/ardrive.ts
```diff
@@ -1002,6 +1013,133 @@ export class ArDrive extends ArDriveAnonymous {
...
+			parentFolderId: folderId,
+			wrappedFile: arweaveManifest,
+			driveId,
+			fileDataRewardSettings,
+			metadataRewardSettings,
+			destFileName: destManifestName,
+			existingFileId
+		});
+
+		const { tipData, reward: communityTipTrxReward } = await this.sendCommunityTip({
```
arielmelendez, line 1116: Consider dry run option

src/ardrive.ts
```diff
@@ -1002,6 +1013,133 @@ export class ArDrive extends ArDriveAnonymous {
...
+					entityId: uploadFileResult.fileId
+				}
+			],
+			tips: [tipData],
+			fees: {
+				[`${uploadFileResult.dataTrxId}`]: uploadFileResult.dataTrxReward,
+				[`${uploadFileResult.metaDataTrxId}`]: uploadFileResult.metaDataTrxReward,
+				[`${tipData.txId}`]: communityTipTrxReward
+			},
+			links: [`arweave.net/${uploadFileResult.dataTrxId}`, ...fileLinks]
```
arielmelendez, line 1139: How did we decide on the name and type of "links"? Just want to make sure I know what other use cases we have in mind for it and when to expect there to be multiple entries here.

src/arfs/arfs_file_wrapper.ts
```diff
@@ -52,8 +52,36 @@ export function wrapFileOrFolder(fileOrFolderPath: FilePath): ArFSFileToUpload |
 export function isFolder(fileOrFolder: ArFSFileToUpload | ArFSFolderToUpload): fileOrFolder is ArFSFolderToUpload {
 	return fileOrFolder instanceof ArFSFolderToUpload;
 }
+export interface ArFSEntityToUpload {
+	gatherFileInfo: () => FileInfo;
+	getFileDataBuffer: () => Buffer;
+	getBaseFileName: () => BaseFileName;
+}
+
+export class ArFSManifestToUpload implements ArFSEntityToUpload {
```
arielmelendez, line 61: Add tests

src/types/ardrive_types.ts
```diff
@@ -148,3 +158,16 @@ export interface GetAllDrivesForAddressParams {
 	address: ArweaveAddress;
 	privateKeyData: PrivateKeyData;
 }
+
+// These manifest interfaces taken from arweave-deploy
+export interface ManifestPathMap {
+	[index: string]: { id: string };
+}
+export interface Manifest {
```
arielmelendez, line 166: Could use a docblock explaining the structure. It's confusing that there's an inner member named `manifest` whose type is actually a hardcoded string.

arielmelendez (changes requested):
Left a few comments about API usability and need for tests.

src/ardrive.ts
fedellen: Dry run seems to be working just fine on `create-manifest`. It is never referred to in the ArDrive methods, only passed down through the constructors.

src/ardrive.ts
fedellen: There will always be multiple entries here, unless used on a completely empty folder or drive (maybe that case should error out?).

As for the name and type of "links", I just added that in there for convenience during testing after taking over Phil's branch. It seems very useful to have links displayed in some form, as users can use the generated manifest links to permanently access their files.

src/ardrive.ts
fedellen: I was originally handling this at the CLI layer, as they are forbidden in conjunction together. But since moving the method back to Core, we should absolutely handle the case where both are provided. 

Added here: 944f9b1

src/ardrive.ts
fedellen: I'm working on an integration test that wraps this whole method before I start this refactor. Currently stuck on stubbing out a `FolderHierarchy`, but I'll be digging more into it tonight/tomorrow.

src/types/ardrive_types.ts
fedellen: Added information from the manifest SPIKE to the some docblocks: 18629ac

src/ardrive.ts
fedellen: Hmm, missed this last comment before. 

Maybe we shouldn't accept both driveID and folderID at the Core level and instead provide that convenience at the CLI layer by gathering the root folderID if a driveID is specified before calling the ArDrive method

src/ardrive.ts
fedellen: Hierarchy is now stubbed! Got an integration test wrapped around this function and added a few refactors that moves this logic elsewhere

src/arfs/arfs_file_wrapper.ts
fedellen: Ah, this file could use some testing too. I'll see what I can do

src/ardrive.ts
fedellen: Dry run is working, its tested using dry run in integration.

src/arfs/arfs_file_wrapper.test.ts
```diff
@@ -0,0 +1,83 @@
...
+	it('getFileDataBuffer function returns a compatible Buffer we can use to upload', () => {
+		const manifest = new ArFSManifestToUpload(stubPublicEntitiesWithPaths, 'TestManifest.json');
+
+		expect(manifest.getFileDataBuffer() instanceof Buffer).to.be.true;
+	});
+});
+
+// describe('ArFSFileToUpload class');
+
+// describe('ArFSFolderToUpload class');
```
fedellen, line 83: Implement these, quick coverage wins

src/arfs/arfsdao.ts
```diff
@@ -612,6 +612,22 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 		);
 	}
 
+	async prepareManifest(
```
fedellen, line 615: This function seems redundant now that the sorting implementation has moved to ArFSManifestToUpload. We should just call these from the ArDrive class

tests/integration/ardrive.int.test.ts
```diff
@@ -860,6 +872,63 @@ describe('ArDrive class - integrated', () => {
 			});
 		});
 	});
+
+	// const stubFolderStructure: ArFSPublicFileOrFolderWithPaths[] =
```
fedellen, line 876: Excise commented code

tests/integration/ardrive.int.test.ts
```diff
@@ -860,6 +872,63 @@ describe('ArDrive class - integrated', () => {
...
+			stub(communityOracle, 'selectTokenHolder').resolves(stubArweaveAddress());
+			stub(arfsDao, 'getPublicNameConflictInfoInFolder').resolves({
+				files: [
+					{
+						fileName: 'CONFLICTING_FILE_NAME',
+						fileId: existingFileId,
+						lastModifiedDate: matchingLastModifiedDate
+					}
+				],
+				folders: [{ folderName: 'CONFLICTING_FOLDER_NAME', folderId: stubEntityID }]
```
fedellen, line 893: We don't test or handle manifest to folder conflict yet

tests/stubs.ts
```diff
@@ -118,12 +122,16 @@ interface StubFileParams {
 	driveId?: DriveID;
 	fileName?: string;
 	txId?: TransactionID;
+	parentFolderId?: FolderID;
+	fileId?: FileID;
```
fedellen, line 126: Implement these parameters on stubPrivateFile to avoid future confusion

src/ardrive.ts
fedellen: I've applied this API change in the latest commits. We only accept a folderID at both layers

tests/integration/ardrive.int.test.ts
fedellen: Handled and tested

src/ardrive.ts
```diff
@@ -1002,6 +1009,84 @@ export class ArDrive extends ArDriveAnonymous {
...
+		const filesAndFolderNames = await this.arFsDao.getPublicNameConflictInfoInFolder(folderId);
+
+		// Manifest becomes a new revision if the destination name conflicts for
+		// --replace and --upsert behaviors, since it will be newly created each time
+		const existingFileId = filesAndFolderNames.files.find((f) => f.fileName === destManifestName)?.fileId;
+		const fileToFolderConflict = filesAndFolderNames.folders.find((f) => f.folderName === destManifestName);
+		if ((existingFileId && conflictResolution === skipOnConflicts) || fileToFolderConflict) {
+			// Return empty results if there is an existing manifest and resolution
+			// is set to skip or if there is a file to folder name conflict
+			return { ...emptyArFSResult, links: [] };
```
arielmelendez, line 1033: Why isn't this an error if there's a folder name conflict?

src/ardrive.ts
arielmelendez: It's looking much better thanks!

src/ardrive.ts
```diff
@@ -1002,6 +1009,84 @@ export class ArDrive extends ArDriveAnonymous {
...
+			existingFileId
+		});
+
+		const { tipData, reward: communityTipTrxReward } = await this.sendCommunityTip({
+			communityWinstonTip: uploadBaseCosts.communityWinstonTip
+		});
+
+		// Setup links array from manifest
+		const fileLinks = Object.keys(arweaveManifest.manifest.paths).map(
+			(path) => `arweave.net/${uploadFileResult.dataTrxId}/${path}`
```
arielmelendez, line 1068: No https://?

src/ardrive.ts
```diff
@@ -1002,6 +1009,84 @@ export class ArDrive extends ArDriveAnonymous {
...
+					entityId: uploadFileResult.fileId
+				}
+			],
+			tips: [tipData],
+			fees: {
+				[`${uploadFileResult.dataTrxId}`]: uploadFileResult.dataTrxReward,
+				[`${uploadFileResult.metaDataTrxId}`]: uploadFileResult.metaDataTrxReward,
+				[`${tipData.txId}`]: communityTipTrxReward
+			},
+			links: [`arweave.net/${uploadFileResult.dataTrxId}`, ...fileLinks]
```
arielmelendez, line 1086: If we add https:// do so here as well.

src/arfs/arfs_file_wrapper.test.ts
```diff
@@ -0,0 +1,160 @@
...
+		// Expect index path to be the first file
+		expect(manifest.manifest.index.path).to.equal('index.html');
+
+		// Assert the structure is consistent with provided stub hierarchy
+		expect(manifest.manifest.paths).to.deep.equal({
+			'file-in-root': {
+				id: '0000000000000000000000000000000000000000000'
+			},
+			'index.html': {
+				id: '0000000000000000000000000000000000000000000'
```
arielmelendez, line 23: Should we use some alternative IDs here to make sure things are ending up in the right slots?

src/arfs/arfs_file_wrapper.test.ts
```diff
@@ -0,0 +1,160 @@
...
+			'parent-folder/child-folder/file-in-child': {
+				id: '0000000000000000000000000000000000000000000'
+			},
+			'parent-folder/file-in-parent': {
+				id: '0000000000000000000000000000000000000000000'
+			}
+		});
+	});
+
+	it('constructs a manifest compatible with arweave gateways', () => {
```
arielmelendez, line 34: How difficult would it be to generate the transaction that we would send and ensure that it's constructed properly? i.e. that the content type is set correctly in the tags.

src/arfs/arfs_file_wrapper.test.ts
```diff
@@ -0,0 +1,160 @@
...
+
+describe('ArFSManifestToUpload class', () => {
+	it('will link to an index.html file in the root of the folder if it exists', () => {
+		const manifest = new ArFSManifestToUpload(stubEntitiesWithPathsAndIndexInRoot, 'DriveManifest.json');
+
+		// Ensure arweave required fields exist
+		expect(manifest.manifest.manifest).to.equal('arweave/paths');
+		expect(manifest.manifest.version).to.equal('0.1.0');
+
+		// Expect index path to be the first file
```
fedellen, line 14:
```suggestion
		// Expect index path to be linked to index.html
```

src/arfs/arfs_file_wrapper.ts
```diff
@@ -52,11 +63,87 @@ export function wrapFileOrFolder(fileOrFolderPath: FilePath): ArFSFileToUpload |
...
+					// Slice off base folder path and the leading "/" so manifest URLs path correctly
+					.slice(baseFolderPath.length + 1)
+					// Replace spaces with underscores for sharing links
+					.replace(/ /g, '_');
+
+				pathMap[path] = { id: `${child.dataTxId}` };
+			}
+		});
+
+		// Use index.html in the specified folder if it exists, otherwise show first file found
```
arielmelendez, line 108: Is this what arkb does for multifile uploads? Just curious

src/arfs/arfs_file_wrapper.ts
```diff
@@ -52,11 +63,87 @@ export function wrapFileOrFolder(fileOrFolderPath: FilePath): ArFSFileToUpload |
...
+		return new ByteCount(Buffer.byteLength(JSON.stringify(this.manifest)));
+	}
+}
 
-export class ArFSFileToUpload {
+export class ArFSFileToUpload implements ArFSEntityToUpload {
 	constructor(public readonly filePath: FilePath, public readonly fileStats: fs.Stats) {
 		if (+this.fileStats.size >= +maxFileSize) {
-			throw new Error(`Files greater than "${maxFileSize}" bytes are not yet supported!`);
+			throw new Error(`Files greater than "${+maxFileSize - 1}" bytes are not yet supported!`);
```
arielmelendez, line 146: What's with the -1? Is maxFileSize not actually the max?

tests/integration/ardrive.int.test.ts
```diff
@@ -75,6 +87,9 @@ describe('ArDrive class - integrated', () => {
 	const unexpectedDriveId = EID(stubEntityIDAlt.toString());
 	const existingFileId = EID(stubEntityIDAlt.toString());
 
+	const matchingLastModifiedDate = new UnixTime(420);
+	const differentLastModifiedDate = new UnixTime(1337);
```
arielmelendez, line 91: 💪

arielmelendez (approved):
Approved with some open questions.

src/arfs/arfs_file_wrapper.ts
fedellen: I found this error message to be imprecise (by 1 byte) when implementing unit tests around it. We're checking if the provided file size is greater than or equal to the max file size variable here.

src/arfs/arfs_file_wrapper.ts
fedellen: They have the same behavior. 

```ts
// arkb buildManifest
   if (Object.keys(paths).includes('index.html')) {
        index = 'index.html';
      } else {
        index = Object.keys(paths)[0];
      }
```

src/ardrive.ts
fedellen: Added: 3aaacf2

src/arfs/arfs_file_wrapper.ts
arielmelendez: Hmm, should it internally be ">" rather than ">="?

src/ardrive.ts
fedellen: Added: ff9cbef

src/arfs/arfs_file_wrapper.test.ts
fedellen: 749e784

src/arfs/arfs_file_wrapper.test.ts
fedellen: Its probably possibly to directly use the arweave library to do this here. I'd have to dig around our arfs methods some more to get started But, it seems seems like check would be at the integration test level so we know that our own methods will apply the content type correctly.

src/arfs/arfs_file_wrapper.ts
fedellen: Actually, checking for greater than or equal to is the imprecise part here. Not the error message.

Updated: 749e784

arielmelendez (approved):

PR #85, opened 2021-11-19
PE-707: Price calc compatibility

fedellen:
This PR removes the shebang from ardriveFactory and removes node-fetch. Both of these issues made Core incompatible with the price calc repo


package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "ardrive-core-js",
-	"version": "1.0.2",
+	"version": "1.0.3",
```
arielmelendez, line 3: So we're just rolling the bump into this branch/release?

fedellen: This branch is off `master` so I've bumped the version here to release, yes. Still in hotfix mode until price calc is working 😬

src/utils/common.ts
```diff
@@ -492,17 +492,17 @@ export async function sanitizePath(path: string): Promise<string> {
...
-	try {
-		const res = await fetch('https://api.coingecko.com/api/v3/simple/price?ids=arweave&vs_currencies=usd');
-		usdPrice = (await res.clone().json()).arweave.usd;
-		return usdPrice;
-	} catch (err) {
-		console.log('Error getting AR/USD price from Coingecko');
-		return 0;
-	}
-}
+// export async function getArUSDPrice(): Promise<number> {
```
arielmelendez, line 495: ?

fedellen: This was using the removed package `node-fetch` and its not being used anywhere so I commented it out for now.

I think we will move the fiat oracles to core for dealing with usd price

arielmelendez (approved):

PR #87, opened 2021-11-22
PE-639: Ask prompt option on uploads

fedellen:
This PR adds support the `--ask` option in ArDrive Core. The growing conflict resolution logic has been moved out of the ArDrive method logic and into its own file.

The Core layer defines all of the types for each prompt to implement, which the CLI (or any other app) will optionally implement.

Suggested review order:

```shell
upload_conflict_resolution.test.ts
upload_conflict_resolution.ts
ardrive.int.test.ts
ardrive.ts
```

### DIFF
--- a/src/types/upload_conflict_types.ts
+++ b/src/types/upload_conflict_types.ts
@@ -1,5 +1,5 @@
 import { FileID, FolderID } from '.';
-import { ArFSFileToUpload, ArFSFolderToUpload } from '../arfs/arfs_file_wrapper';
+import { ArFSEntityToUpload, ArFSFolderToUpload } from '../arfs/arfs_file_wrapper';
 import { FileConflictInfo, NameConflictInfo } from '../utils/mapper_functions';
 
 export const skipOnConflicts = 'skip';
@@ -85,7 +85,7 @@ export interface ResolveNameConflictsParams {
 
 export interface ResolveFileNameConflictsParams extends ResolveNameConflictsParams {
 	destinationFileName: string;
-	wrappedFile: ArFSFileToUpload;
+	wrappedFile: ArFSEntityToUpload;
 	prompts?: FileConflictPrompts;
 }
 
--- a/src/types/upload_conflict_types.ts
+++ b/src/types/upload_conflict_types.ts
@@ -1,0 +1,96 @@
+import { FileID, FolderID } from '.';
+import { FileConflictInfo, NameConflictInfo, ArFSFileToUpload, ArFSFolderToUpload } from '../exports';
+
+export const skipOnConflicts = 'skip';
+export const replaceOnConflicts = 'replace';
+export const upsertOnConflicts = 'upsert';
+export const askOnConflicts = 'ask';
+
+export const renameOnConflicts = 'rename';
+export const useExistingFolder = 'useFolder';
+
+export const errorOnConflict = 'error';
+
+/** Conflict settings used by ArDrive class */
+export type FileNameConflictResolution =
+	| typeof skipOnConflicts
+	| typeof replaceOnConflicts
+	| typeof upsertOnConflicts
+	| typeof askOnConflicts;
+
+export interface ConflictPromptParams {
+	namesWithinDestFolder: string[];
+}
+export interface FileConflictPromptParams extends ConflictPromptParams {
+	fileName: string;
+	fileId: FileID;
+}
+
+export interface FileToFileConflictPromptParams extends FileConflictPromptParams {
+	hasSameLastModifiedDate: boolean;
+}
+
+export interface FolderConflictPromptParams extends ConflictPromptParams {
+	folderName: string;
+	folderId: FolderID;
+}
+
+export type FileToFileNameConflictPrompt = (
+	params: FileToFileConflictPromptParams
+) => Promise<
+	| { resolution: typeof skipOnConflicts | typeof replaceOnConflicts }
+	| { resolution: typeof renameOnConflicts; newFileName: string }
+>;
+
+export type FileToFolderConflictAskPrompt = (
+	params: FolderConflictPromptParams
+) => Promise<{ resolution: typeof skipOnConflicts } | { resolution: typeof renameOnConflicts; newFileName: string }>;
+
+export type FolderToFileConflictAskPrompt = (
+	params: FileConflictPromptParams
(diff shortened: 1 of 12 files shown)


src/types/upload_conflict_types.ts
```diff
@@ -0,0 +1,97 @@
...
+export const useExistingFolder = 'useFolder';
+
+export const errorOnConflict = 'error';
+
+/** Conflict settings used by ArDrive class */
+export type FileNameConflictResolution =
+	| typeof skipOnConflicts
+	| typeof replaceOnConflicts
+	| typeof upsertOnConflicts
+	| typeof askOnConflicts;
```
fedellen, line 20: Consider if and how this effects the new manifest method

src/utils/upload_conflict_resolution.ts
```diff
@@ -0,0 +1,254 @@
...
+
+		return prompts.fileToFileNameConflict({
+			fileId: existingNameAtDestConflict.existingFileConflict.fileId,
+			fileName: destFileName,
+			hasSameLastModifiedDate,
+			namesWithinDestFolder: allExistingNames
+		});
+	})();
+
+	switch (userInput.resolution) {
```
matibat, line 88: Isn't there a default case here?

src/utils/upload_conflict_resolution.ts
```diff
@@ -0,0 +1,254 @@
...
+} from '../types';
+import { NameConflictInfo, FolderNameAndId, FileConflictInfo } from './mapper_functions';
+
+export const resolveFileNameConflicts = async ({
+	wrappedFile,
+	conflictResolution,
+	destinationFileName: destFileName,
+	nameConflictInfo,
+	prompts
+}: ResolveFileNameConflictsParams): Promise<void> => {
```
matibat, line 20: I don't see any references to `this` here. Is this method really needing to be an arrow function?
```suggestion
export async function resolveFileNameConflicts ({
	wrappedFile,
	conflictResolution,
	destinationFileName: destFileName,
	nameConflictInfo,
	prompts
}: ResolveFileNameConflictsParams): Promise<void> {
```

matibat (commented):
Left just a few comments there.
It's looking good and it has very descriptive comments! I'll approve once I dev-test it a little

src/utils/upload_conflict_resolution.ts
fedellen: SGTM: a5d0d6e

matibat (approved):
It's properly working now! Thanks for that quick fix

src/utils/upload_conflict_resolution.ts
```diff
@@ -0,0 +1,254 @@
...
+		if (conflictResolution === replaceOnConflicts) {
+			// Proceed with new revision
+			wrappedFile.existingId = existingNameAtDestConflict.existingFileConflict.fileId;
+			return;
+		}
+
+		// Otherwise, default to upsert behavior
+		if (hasSameLastModifiedDate) {
+			// Skip this file with upsert, it has a matching last modified date
+			wrappedFile.conflictResolution = upsertOnConflicts;
```
arielmelendez, line 53: Comment says skip but returns upsertOnConflicts. And I think the logic should be skipping since the lastModifiedDate is the same.

arielmelendez: FWIW your tests pass because your test case thinks this is the correct behavior:
```
it('resolves wrappedFile.conflictResolution to upsert when there is a file to file name conflict in the destination folder and the file has a MATCHING last modified date and the resolution is set to upsert', async () => {
```

arielmelendez: I think the confusion might be around what it means to upsert. It means to update (if necessary) or insert if it does not exist. So the choice to be made is whether, when we encounter the same last modified date, we consider the files "the same" and NOT in need of an update (i.e. skip). Otherwise we might perform a wasteful upload.

src/utils/upload_conflict_resolution.ts
```diff
@@ -0,0 +1,254 @@
...
+	if (!existingNameAtDestConflict.existingFileConflict && !existingNameAtDestConflict.existingFolderConflict) {
+		// There are no conflicts, continue folder upload
+		return;
+	}
+
+	if (conflictResolution !== askOnConflicts) {
+		if (existingNameAtDestConflict.existingFileConflict) {
+			// Folders cannot overwrite files
+			// Skip this folder and all its contents
+			wrappedFolder.conflictResolution = skipOnConflicts;
```
arielmelendez, line 133: It's interesting that this is a failure case for files unless we're skipping on conflicts, but always a skip for folders. Perhaps it should be an error unless --skip is provided? What's the rationale here? I just worry that a user would either:
• think a folder has uploaded when it hasn't
• not ever realize that a folder didn't upload

arielmelendez (changes requested):
Some open q's re: intended behavior.

jdaev (dismissed):

src/utils/upload_conflict_resolution.ts
fedellen: I separated the case of skipping from the case of skipping during upsert so I could return a different behavior here at the ArDrive level. During a skip, we return an emptyArFSResult. On a skip due to upsert, we return an error to the user explaining that the file is skipped because it matches the file on chain

src/utils/upload_conflict_resolution.ts
fedellen: Throwing an error during a bulk method can result in transaction information being lost to the user. File transactions can be created and uploaded before the error throwing case. The JSON result with transaction IDs would never be sent to the user because of the thrown error

The solution I used here was to never throw errors on the bulk folder method. I was toying around with adding an `errors: []` field to the JSON output as an alternative solution, but it seemed intrusive to the output API.

Now, given the context of glob uploads and errors that could occur during that process, it might be a better idea to go with that `errors` field. What are your thoughts?

arielmelendez: I think it could be worthwhile as long as the exit status is NOT 0. Might be good to see what other common command line tools do in this case. Let's huddle up later to discuss and/or start a Slack thread.

arielmelendez (dismissed):

arielmelendez (approved):

PR #91, opened 2021-11-29
PE-464: download file

matibat:


src/arfs/arfs_entities.ts
```diff
@@ -206,6 +206,8 @@ export class ArFSPrivateFileOrFolderWithPaths extends ArFSFileOrFolderEntity imp
 	}
 }
 
+export type ArFSAnyFileOrFolderWithPaths = ArFSPublicFileOrFolderWithPaths | ArFSPrivateFileOrFolderWithPaths;
```
matibat, line 209: Remove unused code

matibat (commented):
Cleanup needed

src/arfs/arfsdao_anonymous.ts
```diff
@@ -284,4 +288,22 @@ export class ArFSDAOAnonymous extends ArFSDAOType {
 		const entitiesWithPath = children.map((entity) => new ArFSPublicFileOrFolderWithPaths(entity, hierarchy));
 		return entitiesWithPath;
 	}
+
+	/**
+	 * Downloads the data of a public file into certain existing folder in the local storage
+	 * @param file - the file entity to be download
+	 * @returns {Promise<void>}
```
matibat, line 295:
```suggestion
	 * @returns {Promise<{ data: Readable; length: number }>}
```

src/arfs/arfsdao_anonymous.ts
```diff
@@ -284,4 +288,22 @@ export class ArFSDAOAnonymous extends ArFSDAOType {
...
+	 */
+	async getDataStream(file: ArFSPublicFile | ArFSPrivateFile): Promise<{ data: Readable; length: number }> {
+		const fileTxId = file.dataTxId;
+		const dataTxUrl = `${gatewayURL}${fileTxId}`;
```
fedellen, line 299: This could be simplified a bit by just passing a fileId down. Doesn't look like we use any of the other fields within this scope

src/utils/query.ts
```diff
@@ -67,6 +68,7 @@ export function buildQuery({ tags = [], cursor, owner, sort = DESCENDING_ORDER }
 	return {
 		query: `query {
 			transactions(
+				${ids?.length ? `ids: [${ids.map((id) => `"${id}"`)}]` : ''}
```
fedellen, line 71: I'm interested to see if there are any limitations in how many txIds we can pass in here. Its not relevant within the context of download-file though.

src/utils/stream_decrypt.ts
```diff
@@ -0,0 +1,40 @@
+import { createDecipheriv } from 'crypto';
+import { Transform } from 'stream';
+import { CipherIV, FileKey } from '../types';
+
+const algo = 'aes-256-gcm'; // crypto library does not accept this in uppercase. So gotta keep using aes-256-gcm
+const authTagLength = 16;
+
+export class StreamDecrypt extends Transform {
+	// FIXME: holding the data into a buffer uses extra memory. Go back to use disk?
+	private encryptedData: Buffer = Buffer.from(''); // TODO: is it possible to pop the stream data in a buffer-like manner?
```
fedellen, line 10: I'm not entirely sure what these comments mean. So for private downloads, we need to convert it to a Buffer to decrypt, but with public downloads the data can always remain as a stream?

src/arfs/arfs_file_wrapper.ts
```diff
@@ -226,3 +231,53 @@ export class ArFSFolderToUpload {
...
+
+export class ArFSPublicFileToDownload extends ArFSFileToDownload {
+	constructor(fileEntity: ArFSPublicFile, dataStream: Readable, localFilePath: string) {
+		super(fileEntity, dataStream, localFilePath);
+	}
+
+	async write(): Promise<void> {
+		const writeStream = createWriteStream(this.localFilePath); // TODO: wrap 'fs' in a browser-safe class
+		const writePromise = pipelinePromise(this.dataStream, writeStream);
+		writePromise.finally(this.setLastModifiedDate);
```
fedellen, line 264: This seems to be the issue I ran into on the CLI side. We need to await this promise here or return it as the private file `write()` does

src/arfs/arfs_file_wrapper.ts
```diff
@@ -226,3 +231,53 @@ export class ArFSFolderToUpload {
...
+
+export abstract class ArFSFileToDownload {
+	constructor(
+		readonly fileEntity: ArFSFileOrFolderEntity,
+		readonly dataStream: Readable,
+		readonly localFilePath: string
+	) {
+		if (fileEntity.entityType !== 'file') {
+			throw new Error(`Can only download the data of file entities, but got ${fileEntity.entityType}`);
+		}
```
fedellen, line 243: This error case doesn't seem necessary if we just make the abstract fileEntity typesafe with ArFSPublicFile | ArFSPrivateFile

fedellen (changes requested):
Looking good! This cupcake version is much easier to digest. Left comments on mostly minor things and a couple questions. 

Requesting changes because I am getting an error on the CLI side when dev-testing though, might be related to the public file `write()` implementation but I'm not positive

src/arfs/arfsdao_anonymous.ts
arielmelendez: Better yet, could we create an interface for this? e.g. `ReadableData`?

src/arfs/arfs_file_wrapper.ts
matibat: It was a scope issue: References to `this` at setLastModifiedDate were all relative to the promise instead of the `ArFSFileToDownload`. See: 5818313

src/arfs/arfsdao_anonymous.ts
```diff
@@ -284,4 +288,22 @@ export class ArFSDAOAnonymous extends ArFSDAOType {
 		const entitiesWithPath = children.map((entity) => new ArFSPublicFileOrFolderWithPaths(entity, hierarchy));
 		return entitiesWithPath;
 	}
+
+	/**
+	 * Downloads the data of a public file into certain existing folder in the local storage
```
arielmelendez, line 293: Comment needs updating.

src/utils/stream_decrypt.ts
arielmelendez: @matibat @vilenarios and I did some experiments to see if we could get downloading an encrypted file working with a streaming decipher rather than having to queue up all the encrypted bytes. The challenge is that the last 16 bytes of the whole file contain the authTag which is needed to set up the decipher.

We figured out a potential workaround: Using HTTP Range Requests to get the last 16 bytes of the file before trying to do the download, which itself would only request 0-(N-16) bytes. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests for more.

To make use of this, we'll need:
1) A new Dao function to get an auth tag for a dataTx. It will need to compute the full file size of the encrypted data item using the metadata's reported unencrypted size and then put that through the `encryptedDataSize` function. That will give us the upper bound for the range request.
2) Do away with the StreamDecrypt class in favor of simply plugging together the pipes like so:
```
	async downloadPrivateFile(
		fileId: FileID,
		destFolderPath: string,
		driveKey: DriveKey
		// progressCB?: (pctTotal: number, pctFile: number, curFileName: string, curFilePath: string) => void
	): Promise<void> {
		const privateFile = await this.getPrivateFile({ fileId, driveKey });
		const fullPath = joinPath(destFolderPath, privateFile.name);
		const { data } = await this.arFsDao.getDataStream(privateFile);
		const fileKey = await deriveFileKey(`${fileId}`, driveKey);
		const fileCipherIV = await this.arFsDao.getPrivateTransactionCipherIV(privateFile.dataTxId);
		const iv: Buffer = Buffer.from(fileCipherIV, 'base64');
		const decipher = createDecipheriv('aes-256-gcm', fileKey, iv, { authTagLength: 16 });
		const authTag = await this.arFsDao.getAuthTagForDataTxId(privateFile.dataTxId);
		decipher.setAuthTag(authTag);
		const fileToDownload = new ArFSPrivateFileToDownload(privateFile, data, fullPath, decipher);
		await fileToDownload.write();
	}
```

src/ardrive.ts
```diff
@@ -1615,4 +1618,26 @@ export class ArDrive extends ArDriveAnonymous {
...
+		const encryptedDataLength = this.encryptedDataSize(privateFile.size);
+		const { data } = await this.arFsDao.getPrivateDataStream(privateFile.dataTxId, encryptedDataLength);
+		const fileKey = await deriveFileKey(`${fileId}`, driveKey);
+		const fileCipherIV = await this.arFsDao.getPrivateTransactionCipherIV(privateFile.dataTxId);
+		// const iv = Buffer.from(fileCipherIV, 'base64');
+		// const decipher = createDecipheriv('aes-256-gcm', fileKey, iv, { authTagLength: 16 });
+		const authTag = await this.arFsDao.getAuthTagForDataTxId(privateFile.dataTxId, encryptedDataLength);
+		// decipher.setAuthTag(authTag);
+		// Buffer.concat([decipher.update(encryptedDataSlice), decipher.final()]);
+		const decipher = new StreamDecrypt(fileCipherIV, fileKey, authTag);
```
matibat, line 1639: I didn't remove this wrapper class because of some errors regarding lengths and the `authTag`. I'm not sure if it is possible to directly pipe the `DecipherGCM` stream, or even if it's worth it.

It's currently working fine with a simpler implementation.

matibat: TODO: remove these comments

src/arfs/arfs_builders/arfs_file_builders.ts
```diff
@@ -19,10 +19,11 @@ import { ArFSPublicFile, ArFSPrivateFile } from '../arfs_entities';
...
 interface FileMetaDataTransactionData {
+	// FIXME: do we need our safe types here? This interface refers to a JSON with primitive types
 	name: string;
-	size: ByteCount;
-	lastModifiedDate: UnixTime;
-	dataTxId: TransactionID;
+	size: number;
+	lastModifiedDate: number;
+	dataTxId: string;
 	dataContentType: ContentType;
```
matibat, line 27: @fedellen this interface is a type for the parsed JSON data of the Metadata tx. It doesn't actually use our safe types, right?

src/arfs/arfsdao.ts
```diff
@@ -1056,4 +1064,104 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
+			method: 'get',
+			url: dataTxUrl,
+			responseType: 'stream',
+			headers: {
+				Range: `bytes=0-${+authTagIndex - 1}`
+			}
+		};
+		const response = await axios(requestConfig);
+		const { data, headers } = response;
+		const length = new ByteCount(+headers['content-length']); // TODO: remove - it useless if we can compute the size before making the request
```
matibat, line 1136: Remove unused field. Remove unused `ReadableData` type after that

src/arfs/arfsdao.ts
```diff
@@ -1056,4 +1064,104 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
+				method: 'GET',
+				url: `${gatewayURL}${txId}`,
+				headers: {
+					Range: `bytes=${authTagIndex}-${+dataLength - 1}`
+				},
+				responseType: 'stream'
+			}).then((response) => {
+				const { data }: { data: Readable } = response;
+
+				const authTag = Buffer.from('0123456789abcdef');
```
matibat, line 1153: Is this the right way of allocating the buffer in memory? Surely there's a better approach ti instantiate this buff of length 16

src/arfs/arfsdao_anonymous.ts
matibat: **EDIT** no need to return the length of the streamed data
```suggestion
	 * @returns {Promise<Readable>}
```

src/utils/stream_decrypt.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export class StreamDecrypt extends Transform {
+	private readonly decipher: DecipherGCM;
+
+	constructor(cipherIV: CipherIV, fileKey: FileKey, authTag: Buffer) {
+		super();
+		const iv: Buffer = Buffer.from(cipherIV, 'base64');
+		this.decipher = createDecipheriv(algo, fileKey, iv, { authTagLength });
+		this.decipher.setAuthTag(authTag);
+	}
```
matibat, line 16: **for the reviewer**
The decipher is instantiated here

src/utils/stream_decrypt.ts
```diff
@@ -0,0 +1,33 @@
...
+
+	_transform(chunk: Buffer, _encoding: BufferEncoding, next: (err?: Error, data?: Buffer) => void): void {
+		const decrypredChunk = this.decipher.update(chunk);
+		this.push(decrypredChunk);
+		next();
+	}
```
matibat, line 22: **for the reviewer**
The data gets fed to this function and the decrypted data is pushed to the `StreamDecrypt`

src/utils/stream_decrypt.ts
```diff
@@ -0,0 +1,33 @@
...
+
+	_flush(next: (err?: Error, data?: Buffer) => void): void {
+		try {
+			const decipherFinalData = this.decipher.final();
+			this.push(decipherFinalData);
+			next();
+		} catch (err) {
+			next(err);
+		}
+	}
```
matibat, line 32: **for the reviewer**
Finally, when there's no more data to consume the returned buffer of `this.decipher.final()` gets pushed into the `StreamDecrypt`

src/arfs/arfs_builders/arfs_file_builders.ts
arielmelendez: Good catch. It's interesting that the safe types weren't causing errors after JSON.parse? I think the better solution might be to make it a class type and be able to instantiate it from a JSON string, i.e. `FileMetaDataTransactionData.from(json);`

arielmelendez: Let's create a tech debt ticket for this.

src/arfs/arfsdao.ts
```diff
@@ -1056,4 +1064,106 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 			throw new Error(`Invalid password! Please type the same as your other private drives!`);
 		}
 	}
+
+	async getPrivateTransactionCipherIV(txId: TransactionID): Promise<CipherIV> {
+		const results = await this.getCipherIVOfPrivateTransactionIDs([txId]);
+		const fileCipherIvResult = results[0];
```
arielmelendez, line 1070: Should check result count.

arielmelendez (dismissed):

src/arfs/arfs_file_wrapper.ts
matibat: Fix applied by @arielmelendez at 30737770194e7e8fa0077d1f4132bf2aea098413

src/arfs/arfsdao_anonymous.ts
matibat: This change was applied

src/arfs/arfsdao_anonymous.ts
matibat: af70483c0a03d5abcb89d6e63fca0d63e09b0bf0

src/arfs/arfsdao_anonymous.ts
matibat: af70483c0a03d5abcb89d6e63fca0d63e09b0bf0

src/arfs/arfsdao.ts
matibat: 43598f5f4c59b1ba469f44229babc98aa28e234a

src/utils/query.ts
matibat: Hloover suggested keeping it below 100. It might not be the current limit, but he said `I’d keep it 100 for “future-proof”`

matibat: We might cover this case by having a wrapper on top of the `buildQuery` to deal with pagination and the limit of 100 queried entities

src/arfs/arfs_builders/arfs_file_builders.ts
matibat: https://ardrive.atlassian.net/browse/PE-749

src/utils/query.ts
matibat: Here the tech debt ticket: https://ardrive.atlassian.net/browse/PE-750

fedellen (approved):
Looking good!

src/utils/resolve_path.test.ts
```diff
@@ -0,0 +1,6 @@
+describe('resolveLocalFilePath function', () => {
```
jdaev, line 1: Why the empty tests?

jdaev (commented):
Looks solid. Well done.
Left one question.

src/utils/resolve_path.test.ts
matibat: I only need confirmation in the test descriptions to proceed

src/utils/resolve_path.test.ts
```diff
@@ -0,0 +1,6 @@
...
+	it('returns the provided path if the file exists');
```
fedellen, line 2: So now we can provide:

test case 1:  `--dest-folder-name /path/to/existing_file.txt`
test case 2: `--dest-folder-name /path/to/folder` 
test case 3: ` --dest-folder-name /path/to/non_existent_file.txt`

I re-worded the descriptions as I read through the new function so I could break it down and understand them, but yours seem fine as is.

`returns the provided path if a file exists at the destination`
`returns the provided path concatenated with the default file name if a folder exists at the destination`
`returns the parent folder path concatenated with the default file name if the provided path does not exist but the parent directory does`
`throws an error if provided path is invalid`

src/utils/resolve_path.ts
```diff
@@ -0,0 +1,25 @@
...
+			return fullPath;
+		}
+	} catch (e) {
+		const parentFolderPath = dirname(fullPath);
+		const parentPathStats = stat(parentFolderPath);
+		if (parentPathStats.isDirectory()) {
+			// the path does not exist, but the parent is an existing directory
+			return fullPath;
+		}
+		// the path neither its parent exist
```
fedellen, line 22:
```suggestion
		// the path nor its parent folder exist
```

arielmelendez: Technically speaking a `nor` requires a `neither`. So it should be:
```suggestion
		// neither path nor its parent exist
```

src/utils/stream_decrypt.test.ts
```diff
@@ -0,0 +1,41 @@
...
+		let decryptedData = '';
+		decryptingStream.on('data', (chunk: Buffer | string) => {
+			decryptedData = `${decryptedData}${chunk.toString()}`;
+		});
+		return pipelinePromise(mockStream, decryptingStream).then(() => {
+			return expect(decryptedData).to.equal(data);
+		});
+	});
+
+	it('fails while descrypting with wrong secrets', () => {
```
arielmelendez, line 35:
```suggestion
	it('fails while decrypting with wrong secrets', () => {
```

src/arfs/arfsdao.ts
```diff
@@ -1114,8 +1114,8 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 	}
 
 	/**
-	 * Downloads the data of a public file into certain existing folder in the local storage
-	 * @param file - the file entity to be download
+	 * Returns the data stream of a private file
+	 * @param privateFile - the entity of the data to be download
 	 * @returns {Promise<void>}
```
arielmelendez, line 1119:
```suggestion
	 * @returns {Promise<Readable>}
```

arielmelendez (approved):

src/utils/resolve_path.ts
```diff
@@ -0,0 +1,25 @@
+import { statSync } from 'fs';
+import { dirname, join as joinPath, resolve as resolvePath } from 'path';
+
+export function resolveLocalFilePath(destFolderPath: string, defaultFileName: string, stat = statSync): string {
```
arielmelendez, line 4: I still find this function difficult to use correctly:
1) It's called `resolveLocalFilePath` without any explanation about what "resolving" means. An accompanying docblock could possibly help with that.
2) It takes a destFolderPath, but it's a valid use case for that to be a file. That's VERY confusing, especially in the absence of documentation.
3) You HAVE to provide a default file name. So if you know the filename you want to use, then you need to provide it again even though it's not useful.
4) The type for stat is not easily inferred. Could incorrect implementations of stat be provided accidentally?


It's also not clear to me that this logic should live in core. I think it would be simpler if core's download API just took a file path fore where the downloaded file should go. It could then leave decisions about where a file should go, given some arbitrary path, up to the client layer, i.e. the CLI. Think of it this way - why is this logic useful for users of Core? Who needs this functionality generally speaking? What does it have to do with ArFS or ArDrive?

arielmelendez: Also why does `downloadPrivateFile` use this, but downloadPublicFile does not?

arielmelendez: Wouldn't all of this be simpler from the core client's perspective if the public/private download APIs were like:
```
	async downloadPrivateFile(
		fileId: FileID,
		driveKey: DriveKey
		destFolderPath: string,
		destFileName?: string
		// progressCB?: (pctTotal: number, pctFile: number, curFileName: string, curFilePath: string) => void
	): Promise<void> {
```
where an undefined `destFileName` means to use the filename on ArDrive?

arielmelendez: Then all you'd need to do is `assertLocalPathExists` for `destFolderPath`

arielmelendez (changes requested):
I have some concerns with the ease of use of the download APIs as it pertains to paths. I think we have too much logic in core that should instead be in the CLI layer.

src/utils/resolve_path.ts
matibat: I removed the responsibility of core to resolve the file path, that's now the task of CLI. Please review the variable/function/script names, I feel they a little confusing still.

src/utils/resolve_path.test.ts
matibat: Those will surely be moved to the client layer

src/ardrive.ts
```diff
@@ -1615,4 +1610,24 @@ export class ArDrive extends ArDriveAnonymous {
...
+	async downloadPrivateFile(
+		fileId: FileID,
+		driveKey: DriveKey,
+		destFolderPath: string,
+		defaultFileName?: string
+		// progressCB?: (pctTotal: number, pctFile: number, curFileName: string, curFilePath: string) => void
+	): Promise<void> {
+		assertFolderExists(destFolderPath);
+		const privateFile = await this.getPrivateFile({ fileId, driveKey });
+		const outputFileName = defaultFileName ?? privateFile.name;
```
matibat, line 1623: TODO: Ensure the double question mark checks for `undefined` values apart of `null`

src/arfs/arfsdao.ts
```diff
@@ -1056,4 +1062,104 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
+			responseType: 'stream',
+			headers: {
+				Range: `bytes=0-${+authTagIndex - 1}`
+			}
+		};
+		const response = await axios(requestConfig);
+		return response.data;
+	}
+
+	getAuthTagForPrivateFile = async (privateFile: ArFSPrivateFile): Promise<Buffer> =>
```
matibat, line 1137: No `await` being used in this method, the `async` keyword is useless here

matibat: There's no need for the method to be an arrow function, `this` isn't used here

src/arfs/arfsdao.ts
```diff
@@ -1056,4 +1062,104 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
+		new Promise((resolve, reject) => {
```
matibat, line 1138: The promise executor doesn't need to be an arrow function as there's no references to `this` inside

src/arfs/arfsdao.ts
```diff
@@ -1056,4 +1062,104 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
+				const { data }: { data: Readable } = response;
+
+				const authTag = Buffer.alloc(16);
+				let index = 0;
+				data.on('data', (chunk: Buffer) => {
+					authTag.set(chunk, index);
+					index += chunk.length;
+				});
+				data.on('end', () => {
+					resolve(authTag);
```
matibat, line 1158: Should I reject if the `index !== 16` or trust the GW will return the right length?

src/arfs/arfsdao.ts
```diff
@@ -1056,4 +1062,104 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
+				url: `${gatewayURL}${privateFile.dataTxId}`,
+				headers: {
+					Range: `bytes=${authTagIndex}-${+dataLength - 1}`
+				},
+				responseType: 'stream'
+			}).then((response) => {
+				const { data }: { data: Readable } = response;
+
+				const authTag = Buffer.alloc(16);
+				let index = 0;
```
matibat, line 1152: Instead of `index`, use `responseByteCount`

src/arfs/arfsdao.ts
```diff
@@ -1056,4 +1062,104 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
+				method: 'GET',
+				url: `${gatewayURL}${privateFile.dataTxId}`,
+				headers: {
+					Range: `bytes=${authTagIndex}-${+dataLength - 1}`
+				},
+				responseType: 'stream'
+			}).then((response) => {
+				const { data }: { data: Readable } = response;
+
+				const authTag = Buffer.alloc(16);
```
matibat, line 1151: Move the `16` to a constant `AUTH_TAG_LENGTH`

src/utils/resolve_path.ts
```diff
@@ -0,0 +1,25 @@
+import { statSync } from 'fs';
```
matibat, line 1: Remove unused source file

src/utils/resolve_path.test.ts
matibat, line 1: remove unused test file

src/arfs/arfsdao.ts
matibat: Note the const is already defined at `stream_decrypt.ts`; consolidate it

src/utils/common.ts
```diff
@@ -547,3 +548,11 @@ export async function fetchMempool(): Promise<string[]> {
 export function urlEncodeHashKey(keyBuffer: Buffer): string {
 	return keyBuffer.toString('base64').replace('=', '');
 }
+
+/** Computes the size of a private file encrypted with AES256-GCM */
+export function encryptedDataSize(dataSize: ByteCount): ByteCount {
+	if (+dataSize > Number.MAX_SAFE_INTEGER - 16) {
+		throw new Error(`Max un-encrypted dataSize allowed is ${Number.MAX_SAFE_INTEGER - 16}!`);
+	}
+	return new ByteCount((+dataSize / 16 + 1) * 16);
```
matibat, line 557: We could use `bignum` for the `ByteCount` later on

src/ardrive.ts
```diff
@@ -1615,4 +1610,24 @@ export class ArDrive extends ArDriveAnonymous {
...
 		await this.arFsDao.assertValidPassword(password);
 	}
+
+	async downloadPrivateFile(
+		fileId: FileID,
+		driveKey: DriveKey,
+		destFolderPath: string,
+		defaultFileName?: string
+		// progressCB?: (pctTotal: number, pctFile: number, curFileName: string, curFilePath: string) => void
+	): Promise<void> {
```
fedellen, line 1620: This method and downloadUploadFile should be using object parameters. 4 parameters is pretty deep for positional parameters, especially ones that share types like destFolderPath and defaultFileName. 

And all outward facing ArDrive methods were refactored to use this pattern for API stability and extensibility

src/utils/common.ts
arielmelendez: I don't think we'll be working at bignum scale of bytes any time soon 😬

src/ardrive.ts
arielmelendez: From : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator
```
The nullish coalescing operator (??) is a logical operator that returns its right-hand side operand when its left-hand side operand is null or undefined, and otherwise returns its left-hand side operand.
```

src/ardrive_anonymous.ts
```diff
@@ -77,4 +80,19 @@ export class ArDriveAnonymous extends ArDriveType {
 		const children = await this.arFsDao.listPublicFolder({ folderId, maxDepth, includeRoot, owner });
 		return children;
 	}
+
+	async downloadPublicFile(
+		fileId: FileID,
+		destFolderPath: string,
```
arielmelendez, line 86: As @fedellen mentioned elsewhere (perhaps another PR), since we have 2 positional parameters here that the same type and could potentially be reversed accidentally by a caller, we should an object argument with named parameters. The same goes for `downloadPrivateFile`.

src/arfs/arfsdao.ts
arielmelendez: I'd reject if it's the wrong size

arielmelendez (changes requested):
There's a few comments here and an ongoing team discussion about the behaviors of the CLI app given different permutations of folder and file existence.

src/arfs/arfsdao.ts
matibat: 7a273d7e11b85661ce7e59710bd9961931ee48a4

src/arfs/arfsdao.ts
matibat: fd4fa2a8944efb12febaacff65beacefc6ca21dd

src/arfs/arfsdao.ts
matibat: fd4fa2a8944efb12febaacff65beacefc6ca21dd

src/arfs/arfsdao.ts
matibat: 40ceac412c6f19a76b9a92e79342483fb2d68929

src/arfs/arfsdao.ts
matibat: 40ceac412c6f19a76b9a92e79342483fb2d68929

src/utils/resolve_path.ts
matibat: 8a56976d205992b317c6388a6b22b5bac4d0e956

src/utils/resolve_path.test.ts
matibat: 8a56976d205992b317c6388a6b22b5bac4d0e956

src/ardrive.ts
matibat: cbbc3031fc39a49e655e31c1aec31ed24f96bd88

src/ardrive_anonymous.ts
matibat: cbbc3031fc39a49e655e31c1aec31ed24f96bd88

src/utils/resolve_path.test.ts
matibat: These were removed

fedellen (dismissed):

src/ardrive.ts
```diff
@@ -1608,13 +1609,12 @@ export class ArDrive extends ArDriveAnonymous {
...
-		destFolderPath: string,
-		defaultFileName?: string
-		// progressCB?: (pctTotal: number, pctFile: number, curFileName: string, curFilePath: string) => void
-	): Promise<void> {
+	async downloadPrivateFile({
+		fileId,
+		driveKey,
+		destFolderPath,
+		defaultFileName
+	}: DownloadPrivateFileArguments): Promise<void> {
```
arielmelendez, line 1617: Naming nit, but I think we need standards around this: Input names in the interface definition are "Parameters". Arguments are the variables/values that you pass in from the outside. The in-function references are parameters. Everywhere else in the codebase we refer to them as Parameters or Params so let's stick with the convention.

src/arfs/arfsdao.ts
```diff
@@ -1134,10 +1140,10 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 		return response.data;
 	}
 
-	getAuthTagForPrivateFile = async (privateFile: ArFSPrivateFile): Promise<Buffer> =>
-		new Promise((resolve, reject) => {
+	getAuthTagForPrivateFile(privateFile: ArFSPrivateFile): Promise<Buffer> {
+		return new Promise(function (resolve, reject) {
```
arielmelendez, line 1144: Is there a reason we're creating a promise and using resolve/reject vs. simply returning or throwing? I don't see an async context here where we're trying to catch an exception.

src/ardrive.ts
arielmelendez: 2b7d7e4

arielmelendez (dismissed):

src/arfs/arfsdao.ts
matibat: Simplified the function by asking Axios to return a buffer: f217cf8a482423a70d1655f2507029f4843d2cfc

arielmelendez (dismissed):

fedellen (dismissed):

fedellen (approved):

arielmelendez (approved):

fedellen (approved):

arielmelendez (approved):

PR #93, opened 2021-11-30
fix(create manifest): URL encode links for user output PE-743

fedellen:


arielmelendez (dismissed):

arielmelendez (changes requested):
Need an error case in core for an empty folder and a matching integration test.

arielmelendez (approved):

jdaev (approved):

PR #95, opened 2021-12-02
refactor(create manifest): Add manifest field to result PE-740

fedellen:
This PR adds the `manifest` field to the `ArFSManifestResult` which gives users access to their manifest `.json` file. They can then pipe a copy of the manifest or verify the structure  with a `--dry-run `

### DIFF
--- a/src/types/ardrive_types.ts
+++ b/src/types/ardrive_types.ts
@@ -38,6 +38,7 @@ export interface ArFSResult {
 }
 
 export interface ArFSManifestResult extends ArFSResult {
+	manifest: Manifest;
 	links: string[];
 }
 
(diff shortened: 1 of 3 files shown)


src/types/ardrive_types.ts
```diff
@@ -47,6 +48,16 @@ export const emptyArFSResult: ArFSResult = {
 	fees: {}
 };
 
+export const emptyManifestResult: ArFSManifestResult = {
+	...emptyArFSResult,
+	manifest: {
```
arielmelendez, line 53: Would it make more sense to just have an empty object here?

arielmelendez (approved):
Looks good. Some enhancements to readme re: manifests might be good too.

matibat (approved):
Very cupcacke-ish!

jdaev (approved):
LGTM

PR #97, opened 2021-12-06
feat: temporary fixes wrong types PE-462

matibat:


src/arfs/arfs_builders/arfs_drive_builders.ts
```diff
@@ -189,7 +189,7 @@ export class ArFSPrivateDriveBuilder extends ArFSMetadataEntityBuilder<ArFSPriva
 			const decryptedDriveJSON: DriveMetaDataTransactionData = await JSON.parse(decryptedDriveString);
 
 			this.name = decryptedDriveJSON.name;
-			this.rootFolderId = decryptedDriveJSON.rootFolderId;
+			this.rootFolderId = EID(decryptedDriveJSON.rootFolderId);
```
jdaev, line 192: Maybe consider a more informative name than EID?

jdaev (approved):
Looks good to me.

fedellen (approved):

src/arfs/arfs_builders/arfs_drive_builders.ts
arielmelendez: That was a choice made by me a while back for a convenience function to do `new EntityID(someIdSt)`. It's a pattern in the codebase. What we could use is just a docblock over the function to explain that to code browsers.

src/arfs/arfs_builders/arfs_drive_builders.ts
```diff
@@ -314,9 +314,10 @@ export class SafeArFSDriveBuilder extends ArFSMetadataEntityBuilder<ArFSDriveEnt
 				if (isPrivate) {
 					// Type-check private properties
 					if (this.cipher?.length && this.driveAuthMode?.length && this.cipherIV?.length) {
+						const rootFolderId: unknown = new EncryptedEntityID(); // FIXME: https://ardrive.atlassian.net/browse/PE-749
```
arielmelendez, line 317: Should we just interpolate `new EncryptedEntityID()` to a string instead? i.e.
```
const rootFolderId = `${new EncryptedEntityID()}`;
```

matibat:
This work is no longer needed. The upcoming ticket: PE-749 will handle these cases

PR #99, opened 2021-12-06
PE-753 - ArFSEntityCache

arielmelendez:
Opening up a draft PR for this to share architecture and implementation possibility with the team.


src/arfs/arfsdao.ts
```diff
@@ -694,7 +706,16 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 
 	// Convenience function for known-private use cases
 	async getPrivateDrive(driveId: DriveID, driveKey: DriveKey, owner: ArweaveAddress): Promise<ArFSPrivateDrive> {
-		return new ArFSPrivateDriveBuilder({ entityId: driveId, arweave: this.arweave, key: driveKey, owner }).build();
+		const cacheKey = { driveId, driveKey, owner };
```
matibat, line 709: Casting this to a string is `'[object Object]'`

src/arfs/ArFSEntityCache.ts
```diff
@@ -0,0 +1,27 @@
+import { Cache, SimpleCache } from '@alexsasharegan/simple-cache';
+
+export class ArFSEntityCache<K, V> {
+	private cache: Cache<string, Promise<V>>;
+
+	constructor(capacity: number) {
```
matibat, line 6: A default here might be useful

src/arfs/ArFSEntityCache.ts
```diff
@@ -0,0 +1,27 @@
...
+		this.cache = SimpleCache<string, Promise<V>>(capacity);
+	}
+
+	put(key: K, value: Promise<V>): Promise<V> {
+		this.cache.write(`${key}`, value);
```
matibat, line 11: Does the library support promises? Would it cache the promise's result or the promise instance itself?
Maybe we should await for the value here

matibat (commented):
I liked this idea, seems very simple. It would be nice to test the `ArFSEntityCache` with some of the safe types apart from the primitives

Left some comments

src/arfs/arfsdao.ts
arielmelendez: Wow good catch. Let me write some tests to catch that.

src/arfs/ArFSEntityCache.ts
arielmelendez: I'd like the implementors to think carefully about what the caching needs might be in their intended use case.

src/arfs/ArFSEntityCache.ts
arielmelendez: It depends what you mean by "support promises". It supports them in that it will accept them as value types to cache. And when it caches them it caches the actual promise, not the result. This allows multiple callers to await on the result if it's taking a long time to come in. Having put some diagnostic logging code in along the way, I've verified this to work as expected. It's similar to the `setupPromise` concept we've used elsewhere (i.e. price calc).

arielmelendez: We can't await for the value or it will cause cache misses in other contexts while the request is in flight. By caching promises, we can make this cache double as a request AND a result cache.

src/arfs/arfsdao.ts
arielmelendez: f079ac1

fedellen (dismissed):
Awesome updates!

matibat (dismissed):
It looks good!

arielmelendez:
Merging with @agsuy 's slack blessing.

PR #101, opened 2021-12-07
PE-763: Dependency management for bundles

fedellen:
This PR adds `arbundles`, bumps `arweave` to be compatible with `arbundles`, and removes `arweave-bundles` to prepare `ardrive-core-js` for bundle compatibility


jdaev (approved):
Looks good 👍🏽

arielmelendez (approved):

src/community/smartweave_contract_oracle.ts
```diff
@@ -12,6 +12,8 @@ export class SmartweaveContractReader implements ContractReader {
 
 	/** Fetches smartweave contracts from Arweave with smartweave-js */
 	async readContract(txId: TransactionID): Promise<unknown> {
+		// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+		// @ts-ignore
```
matibat, line 16: What are we ignoring here?

matibat (approved):
It looks good!
Left one single Q

src/community/smartweave_contract_oracle.ts
fedellen: Its silencing a type error. Smartweave is expecting an arweave `1.10.16`, but we now have version `1.10.18`.  Updating smartweave does not seem to resolve the error and the package is still working.

PR #102, opened 2021-12-07
PE-458: Create bundled drives

fedellen:
This PR adds support for direct to network bundles on the create drive methods.


src/ardrive.ts
```diff
@@ -1145,14 +1135,13 @@ export class ArDrive extends ArDriveAnonymous {
 			reward: driveUploadCosts.rootFolderMetaDataBaseReward,
 			feeMultiple: this.feeMultiple
 		};
-		const createDriveResult = await this.arFsDao.createPublicDrive(
+		const createDriveResult = await this.arFsDao.createPublicDrive({
 			driveName,
 			driveRewardSettings,
```
fedellen, line 1140: TODO on this PR: Update estimate and assert method to accurately account for bundle. 

The estimation is working correctly right now because of our bundle here will always be one chunk, but that is reliant upon pricing to always be chunk based.

src/ardrive.ts
```diff
@@ -1216,11 +1225,32 @@ export class ArDrive extends ArDriveAnonymous {
...
+			});
+
+			// Bundled trx result:
+			return {
+				...arFSResults,
+				fees: {
+					[`${createDriveResult.bundleTrxId}`]: createDriveResult.bundleTrxReward
+				}
+			};
+		}
```
fedellen, line 1244: TODO on this PR: Would like to add `createDrive` function similar to the pattern in ArFSDAO that shares all of this repeated logic

src/arfs/arfs_entity_result_factory.ts
```diff
@@ -1,18 +1,35 @@
 import { ArFSFileMetadataTransactionData } from './arfs_trx_data_types';
 import { DriveID, FolderID, FileID, FileKey, DriveKey, TransactionID, Winston } from '../types';
 
+export interface ArFSBundleWriteResult {
+	bundleTrxId: TransactionID;
+	bundleTrxReward: Winston;
+	metaDataTrxId: TransactionID;
```
fedellen, line 7: TODO on this PR: Update all `Trx` named variable to use the `Tx` standard

src/types/ardrive_types.ts
```diff
@@ -104,8 +105,9 @@ export interface CreatePublicManifestParams extends Required<UploadPublicManifes
 
 export interface CreatePublicFolderParams {
 	folderName: string;
-	driveId: DriveID;
 	parentFolderId: FolderID;
+	/** @deprecated ArFS cache makes passing driveId here redundant. This parameter makes the api confusing and will no longer used */
+	driveId?: DriveID;
```
fedellen, line 110: Note for the reviewer: 

Accepting driveId AND parentFolderId here made the API confusing and forced us to unnecessarily check that the provided parentFolderId does indeed belong to the specified drive. 

One reason for driveId being passed down here was not waste a network call in the CLI layer while deriving the driveKey. The upcoming ArFS cache PR makes this a non-factor, and therefore driveId has no purpose here anymore.

Because this is a public function on the ArDrive class that library consumers could be depending on, I've taken extra precaution in here by making it optional and deprecating the parameter rather than outright removal

src/arfs/arfs_prototypes.ts
```diff
@@ -44,31 +45,35 @@ export abstract class ArFSObjectMetadataPrototype {
 
 export abstract class ArFSEntityMetaDataPrototype extends ArFSObjectMetadataPrototype {
 	readonly unixTime: UnixTime;
+	abstract readonly contentType: ContentType;
```
fedellen, line 48: Note for the reviewer:

We now have coverage for all of these changes (and more) in `arfsdao.test.ts`

src/arfs/arfs_entity_result_factory.ts
fedellen: 4967ac2

src/pricing/arfs_cost_estimator.test.ts
```diff
@@ -0,0 +1,118 @@
...
+					await privateCreateDriveParams()
+				);
+				const v2RewardSettings = rewardSettings as CreateDriveV2TxRewardSettings;
+
+				expect(+totalWinstonPrice).to.equal(6);
+				expect(+v2RewardSettings.driveRewardSettings.reward!).to.equal(3);
+				expect(+v2RewardSettings.driveRewardSettings.feeMultiple!).to.equal(1);
+				expect(+v2RewardSettings.rootFolderRewardSettings.reward!).to.equal(3);
+				expect(+v2RewardSettings.rootFolderRewardSettings.feeMultiple!).to.equal(1);
+			});
```
fedellen, line 115: These expectations do not change from the public test. Might not be worth testing private expectations vs public

tests/integration/ardrive.int.test.ts
```diff
@@ -1099,11 +1136,12 @@ describe('ArDrive class - integrated', () => {
 function assertCreateDriveExpectations(
 	result: ArFSResult,
 	driveFee: Winston,
-	folderFee: Winston,
-	expectedDriveKey?: string
+	folderFee?: Winston,
+	expectedDriveKey?: string,
+	isBundled = false
 ) {
 	// Ensure that 2 arfs entities were created
```
fedellen, line 1143: Update comment

src/ardrive.ts
```diff
@@ -1105,122 +1114,176 @@ export class ArDrive extends ArDriveAnonymous {
...
 			],
 			tips: [],
 			fees: {
-				[`${metaDataTrxId}`]: metaDataTrxReward
+				[`${metaDataTxId}`]: metaDataTxReward
 			}
 		});
 	}
 
+	private readonly costEstimator: ArFSCostEstimator = new ArFSCostEstimator({
```
fedellen, line 1145: Move this to the top of the class declaration. Or into the constructor

fedellen: It should be injectable

src/ardrive.ts
```diff
@@ -1105,122 +1114,176 @@ export class ArDrive extends ArDriveAnonymous {
...
+		};
 	}
 
 	public async createPrivateDrive({ driveName, newPrivateDriveData }: CreatePrivateDriveParams): Promise<ArFSResult> {
-		// Assert that there's enough AR available in the wallet
-		const stubRootFolderData = await ArFSPrivateFolderTransactionData.from(driveName, newPrivateDriveData.driveKey);
-		const stubDriveData = await ArFSPrivateDriveTransactionData.from(
-			driveName,
+		// Use stub data to estimate costs since actual data requires entity IDs generated by ArFSDao
+		const rootFolderMetaDataPrototype = new ArFSPrivateFolderMetaDataPrototype(
```
fedellen, line 1220: Maybe these prototypes should be implementation details for the `ArFSCostEstimator` itself. We could just hand `estimateCreateDrive` a driveName and driveKey for this use case. ArDrive would then be fully separated of the ArFS Prototype details

src/arfs/arfs_prototypes.ts
```diff
@@ -151,14 +150,12 @@ export class ArFSPrivateFolderMetaDataPrototype extends ArFSFolderMetaDataProtot
...
-	}
-
-	addTagsToTransaction(transaction: Transaction): void {
-		super.addTagsToTransaction(transaction);
-		transaction.addTag('Cipher', this.objectData.cipher);
-		transaction.addTag('Cipher-IV', this.objectData.cipherIV);
+	get gqlTags(): GQLTagInterface[] {
+		return [
+			...super.gqlTags,
+			{ name: 'Cipher', value: this.objectData.cipher },
```
jdaev, line 156: Wouldn't it be better to put the tag names in a constants file?

src/ardrive.ts
fedellen: Added `estimation_prototypes.ts` for this. `ArFSCostEstimator` didn't need the extra logic.

I've been wondering whether we need estimation prototype at all though. The only blocked seems to be having access to the UUIDs being created in the `ArFSDAO` layer.

But perhaps this doesn't need to be the case. We could have the `ArFSMetadataPrototype` class optionally take an `existingId` and generate a new one.  The we just check for an existing ID, and construct the REAL ArFSPrototype with a real ID to both estimate and also to pass down to ArFSDAO.

Just a thought,  not sure if I'll try that refactor out.

src/types/ardrive_types.ts
```diff
@@ -14,16 +14,17 @@ import {
...
+export type ArFSEntityDataType = 'drive' | 'folder' | 'file' | 'bundle';
 
 export interface ArFSEntityData {
 	type: ArFSEntityDataType;
-	metadataTxId: TransactionID;
+	bundleTxId?: TransactionID;
+	metadataTxId?: TransactionID;
 	dataTxId?: TransactionID;
-	entityId: AnyEntityID;
+	entityId?: AnyEntityID;
```
matibat, line 27: Why is this guy nullable now?

matibat (commented):
I'm glad to see that many tests there!
Left a couple of questions. A little dev-testing is pending on my side to approve this one

src/ardrive.ts
```diff
@@ -97,11 +99,24 @@ export class ArDrive extends ArDriveAnonymous {
...
 		super(arFsDao);
+
+		this.costEstimator =
+			injectedCostEstimator ??
+			new ArFSCostEstimator({
+				bundle: this.bundle,
+				baseTags: this.arFsDao.baselineArFSTags,
+				feeMultiple: this.feeMultiple,
+				priceEstimator: this.priceEstimator
+			});
```
fedellen, line 115: This part needs to resolved a different way before we move on with this PR. Working constructor params on the ArDrive class needs to be handled carefully and agreed upon. I'm working on an ArFSTagBuilder PR that will merge into this

src/types/ardrive_types.ts
fedellen: Because `bundle` type will not have an entity ID

src/arfs/arfsdao.test.ts
```diff
@@ -1,15 +1,31 @@
...
-import { ArFSPublicFileMetadataTransactionData } from './arfs_trx_data_types';
+
 import { expect } from 'chai';
-import { Tag } from 'arweave/node/lib/transaction';
+import { expectAsyncErrorThrow, getDecodedTags } from '../../tests/test_helpers';
+import { deriveFileKey, driveDecrypt, fileDecrypt } from '../utils/crypto';
+import { DataItem } from 'arbundles';
 
-describe('The ArFSDAO class', () => {
+describe('The ArFSDAO class', async () => {
```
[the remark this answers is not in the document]
fedellen: Doesn't need to be anymore, good catch! Earlier in the commit stream I was constructing all of the private stubs within this scope which are promises. Thats factored out now, will excise

src/arfs/arfs_prototypes.ts
fedellen: Yeah, having constants for these would help ensure accuracy all over the code base. I can do that refactor pretty quickly, but it will be a huge effect on the diff. Another ticket is best: [PE-800](https://ardrive.atlassian.net/browse/PE-800)

src/arfs/arfs_tag_settings.ts
```diff
@@ -0,0 +1,118 @@
+import { GQLTagInterface, TipType } from '../types';
+import { CURRENT_ARFS_VERSION, DEFAULT_APP_NAME, DEFAULT_APP_VERSION } from '../utils/constants';
+
+// Tag Limits to be in compliance with ANS-104:
+// https://github.com/joshbenaron/arweave-standards/blob/ans104/ans/ANS-104.md#21-verifying-a-dataitem
```
arielmelendez, line 5: Might want to lock in this URL at a particular git revision.

arielmelendez (approved):

src/ardrive_factory.ts
```diff
@@ -16,11 +16,14 @@ import { ArFSTagSettings } from './arfs/arfs_tag_settings';
...
 export interface ArDriveSettingsAnonymous {
 	arweave?: Arweave;
+	/** @deprecated App Version is an unused parameter on anonymous ArDrive */
 	appVersion?: string;
+	/** @deprecated App Name is an unused parameter on anonymous ArDrive */
 	appName?: string;
-	arFSTagSettings?: ArFSTagSettings;
 }
 export interface ArDriveSettings extends ArDriveSettingsAnonymous {
+	appVersion?: string;
```
arielmelendez, line 25: Perhaps we should opportunistically deprecate this as well?

fedellen: Sure, it will make more sense in a future major version to just provide a tag settings class

arielmelendez (dismissed):

matibat (commented):
It's looking good to me! Gotta dev-test it a little before approving

arielmelendez (dismissed):

matibat (approved):
It properly works on my end!

PR #106, opened 2021-12-14
PE-458: ArFS Tag Settings

fedellen:
This PR adds the `ArFSTagBuilder` class which will now be responsible for handling base ArFS tx tags, tip tx tags, and bundle tx tags. It comes wrapped in its own unit tests and is used within the ArFSDAO and ArDrive classes.

The tag builder is meant to solve a circular dependency problem with passing base tags around. Estimating dataItems does require knowing the exact character length of every tag. This means that we need all of the tags fully constructed within the ArDrive layer to accurately estimate and assert an action. We use this calculation passed down as a `rewardSetting` to avoid a network request price check during transaction creation.

The refactor here causes `appVersion` and `appName` to be deprecated fields within the ArDrive and ArFSDAO class constructors. We will deprecate for now to maintain backwards compatibility, and during the next major version bump we can re-evaluate and re-structure our constructor params. The updated ArDrive Factory remains the ideal way to construct all classes with simple params

Suggested review order:

`arfs_tag_builder.test.ts`
`arfs_tag_builder.ts`
`arfsdao.ts` / `ardrive.ts` constructors


src/ardrive.test.ts
```diff
@@ -42,17 +44,23 @@ describe('ArDrive class', () => {
 		arweaveOracleStub.getWinstonPriceForByteCount.callsFake((input) => Promise.resolve(W(+input)));
 		communityOracleStub = stub(new ArDriveCommunityOracle(fakeArweave));
 		priceEstimator = new ARDataPriceRegressionEstimator(true, arweaveOracleStub);
-		walletDao = new WalletDAO(fakeArweave, 'Unit Test', '1.0');
+		walletDao = new WalletDAO(fakeArweave, 'Unit Test', '1.2');
+
+		const arFSTagBuilder = new ArFSTagBuilder('Unit Test', '1.2');
+		const costEstimator = new ArFSCostEstimator({ arFSTagBuilder, priceEstimator });
+		// const bundledCostEstimator = new ArFSCostEstimator({ bundle: false, arFSTagBuilder, priceEstimator });
```
fedellen, line 51: Excise

src/ardrive.ts
```diff
@@ -88,35 +86,31 @@ import {
...
+		// @ts-ignore
 		private readonly appName: string,
+		/** @deprecated appVersion is now always read from ArFSTagBuilder */
+		// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+		// @ts-ignore
 		private readonly appVersion: string,
 		private readonly priceEstimator: ARDataPriceEstimator = new ARDataPriceChunkEstimator(true),
 		private readonly feeMultiple: FeeMultiple = new FeeMultiple(1.0),
 		private readonly dryRun: boolean = false,
-		private readonly bundle: boolean = true,
```
fedellen, line 103: Good to note here for the reviewer that the `bundle` setting now exclusively lives in the `ArFSCostEstimator`. We infer whether to bundle or not bases on how we estimated the cost of the tx

src/arfs/arfs_tag_builder.test.ts
```diff
@@ -0,0 +1,80 @@
...
+			).to.deep.equal([
+				{ name: 'App-Name', value: 'Tag-Builder-Test' },
+				{ name: 'App-Version', value: '1.2' },
+				{ name: 'ArFS', value: '0.101' },
+				{ name: 'Custom-Tag-1', value: 'Gibberish' },
+				{ name: 'Custom-Tag-2', value: 'Excited penguin' }
+			]);
+		});
+
+		it('returns provided tags combined with the base arfs tags', () => {
```
fedellen, line 49: Update test desc

src/arfs/arfs_tag_builder.test.ts
```diff
@@ -0,0 +1,80 @@
...
+				{ name: 'App-Name', value: 'Tag-Builder-Test' },
+				{ name: 'App-Version', value: '1.2' },
+				{ name: 'Bundle-Format', value: 'binary' },
+				{ name: 'Bundle-Version', value: '2.0.0' },
+				{ name: 'Custom-Tag-5', value: 'Monkey paradise' },
+				{ name: 'Custom-Tag-8', value: 'Mountain sun' }
+			]);
+		});
+
+		it('can exclude any specified tags names ', () => {
```
fedellen, line 73: Whitespace at end

src/arfs/arfs_tag_builder.test.ts
```diff
@@ -0,0 +1,80 @@
...
+				{ name: 'App-Name', value: 'Tag-Builder-Test' },
+				{ name: 'App-Version', value: '1.2' },
+				{ name: 'ArFS', value: '0.101' },
+				{ name: 'Custom-Tag-1', value: 'Gibberish' },
+				{ name: 'Custom-Tag-2', value: 'Excited penguin' }
+			]);
+		});
+
+		it('can exclude any specified tags names', () => {
+			expect(arFSTagBuilder.withBaseArFSTags([], ['App-Version', 'ArFS'])).to.deep.equal([
```
arielmelendez, line 50: Seeing this API used this way, it makes me think that object params might be beneficial so that the excludes aren't visually confused with anything else. Another option could be to provide them at construction time of the builder.

arielmelendez: To be clear: I understand the type checking would prevent any issues (even in the case of empty arrays), but I think readability could be enhanced.

src/arfs/arfs_tag_builder.test.ts
```diff
@@ -0,0 +1,80 @@
+import { expect } from 'chai';
+import { ArFSTagBuilder } from './arfs_tag_builder';
+
+describe('ArFSTagBuilder class', () => {
```
arielmelendez, line 4: Most "Builder" classes' APIs allow you to pump data in and then call a `build()` function which would potentially through an error if any validation checks fail. This class is more of a `Helper`. Consider a name change.

src/arfs/arfs_tag_builder.ts
```diff
@@ -0,0 +1,45 @@
+import { CURRENT_ARFS_VERSION, GQLTagInterface, TipType } from '../types';
+
+export class ArFSTagBuilder {
+	constructor(
+		private readonly appName: string,
```
arielmelendez, line 5: Consider object params as all three of these are easily interchangeable.

src/arfs/arfs_tag_builder.ts
```diff
@@ -0,0 +1,45 @@
+import { CURRENT_ARFS_VERSION, GQLTagInterface, TipType } from '../types';
+
+export class ArFSTagBuilder {
```
arielmelendez, line 3: We likely have a responsibility to throw an exception when any of the tag combos is out of spec. Here's the latest from the ANS-104 spec (https://github.com/joshbenaron/arweave-standards/blob/ans104/ans/ANS-104.md#21-verifying-a-dataitem):
```
A tag object is valid iff.:
• there are <= 128 tags
• each key is <= 1024 bytes
• each value is <= 3072 bytes
• only contains a key and value
• both the key and value are non-empty strings
```
We'll need to enforce these rules, with some new types and/or new logic here.

src/ardrive.ts
arielmelendez: That's not so intuitive based on the class name. Let's strategize about that.

src/ardrive.ts
```diff
@@ -88,35 +86,31 @@ import {
...
 } from './pricing/estimation_prototypes';
+import { ArFSTagBuilder } from './arfs/arfs_tag_builder';
 
 export class ArDrive extends ArDriveAnonymous {
 	constructor(
 		private readonly wallet: Wallet,
 		private readonly walletDao: WalletDAO,
 		protected readonly arFsDao: ArFSDAO,
 		private readonly communityOracle: CommunityOracle,
+		/** @deprecated appName is now always read from ArFSTagBuilder */
```
arielmelendez, line 97: Note: If it's always read from the ArFSTagBuilder, and if the tag builder and cost estimator are required, then we've effectively broken this API and it is no longer backwards compatible. Is it possible to add this architectural change without disrupting backwards compatibility?

src/arfs/arfs_tag_builder.test.ts
arielmelendez: Or maybe ArFSTagSettings?

src/arfs/arfsdao.ts
```diff
@@ -141,10 +142,17 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
+		/** @deprecated appName is now always read from ArFSTagBuilder */
+		// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+		// @ts-ignore
 		protected appName = DEFAULT_APP_NAME,
-		protected appVersion = DEFAULT_APP_VERSION
+		/** @deprecated appVersion is now always read from ArFSTagBuilder */
+		// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+		// @ts-ignore
+		protected appVersion = DEFAULT_APP_VERSION,
+		protected readonly arFSTagBuilder: ArFSTagBuilder
```
arielmelendez, line 153: Same comment here about breaking backwards compatibility.

src/arfs/arfsdao.ts
```diff
@@ -678,31 +675,23 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
+		const tags: GQLTagInterface[] = this.arFSTagBuilder.withBaseBundleTags(otherTags, excludedTagNames);
 
 		// If we've opted to boost the transaction, do so now
 		if (rewardSettings.feeMultiple?.wouldBoostReward()) {
 			bundledDataTx.reward = rewardSettings.feeMultiple.boostReward(bundledDataTx.reward);
 
-			// Add a Boost tag
-			tags.push({ name: 'Boost', value: rewardSettings.feeMultiple.toString() });
+			// Add a Boost tag if not excluded
+			if (!excludedTagNames.includes('Boost')) {
```
arielmelendez, line 685: Oooh that's a tricky case.

arielmelendez: Could it make more sense to push it onto "otherTags" and then use the tags builder down below and let it get filtered out by `withBaseBundleTags`?

src/arfs/arfsdao.ts
```diff
@@ -714,7 +703,8 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 	}: ArFSPrepareObjectTransactionParams): Promise<Transaction> {
 		// Enforce that other tags are not protected
 		objectMetaData.assertProtectedTags(otherTags);
-		const tags = [...this.baselineArFSTags, ...objectMetaData.gqlTags, ...otherTags];
+
+		const tags = this.arFSTagBuilder.withBaseArFSTags([...objectMetaData.gqlTags, ...otherTags], excludedTagNames);
```
arielmelendez, line 707: Same comments as above about moving this down below the scope of the boost checks.

src/arfs/arfsdao_anonymous.ts
```diff
@@ -41,8 +42,15 @@ export abstract class ArFSDAOType {
...
+		/** @deprecated appName is now always read from ArFSTagBuilder */
+		// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+		// @ts-ignore
 		protected appName = DEFAULT_APP_NAME,
-		protected appVersion = DEFAULT_APP_VERSION
+		/** @deprecated appVersion is now always read from ArFSTagBuilder */
+		// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+		// @ts-ignore
+		protected appVersion = DEFAULT_APP_VERSION,
+		protected readonly arFSTagBuilder: ArFSTagBuilder
```
arielmelendez, line 53: Same comments about backwards compatibility.

src/types/cost_estimator_types.ts
```diff
@@ -17,6 +18,7 @@ export interface EstimateCreateDriveParams {
 export interface EstimateResult<T> {
 	totalWinstonPrice: Winston;
 	rewardSettings: T;
+	// TODO: Add Bundle Plan { numOfBundles: number , ... }
```
arielmelendez, line 21: Is this still todo?

src/ardrive_factory.ts
```diff
@@ -37,36 +41,45 @@ const defaultArweave = Arweave.init({
...
-	dryRun,
-	feeMultiple,
-	arfsDao,
+	dryRun = false,
+	feeMultiple = new FeeMultiple(1.0),
 	appName = DEFAULT_APP_NAME,
 	appVersion = DEFAULT_APP_VERSION,
-	bundle = true
+	walletDao = new WalletDAO(arweave, appName, appVersion),
+	bundle = true,
```
arielmelendez, line 53: bundle sounds like a noun. Consider `shouldBundle`

arielmelendez (changes requested):
Left a number of important comments. Let's circle up if you have questions or concerns.

src/ardrive.ts
fedellen: Hmm. We could probably achieve this by making the new fields optional and use the constructor to supply the defaults. That way we have access to the appVersion/appName 

I'll try some things out

src/ardrive.ts
fedellen: Haven't renamed the `ArFSCostEstimator` class yet. Maybe `ArFSWriteOperationPreparer` or similar?

src/ardrive_factory.ts
```diff
@@ -37,36 +41,45 @@ const defaultArweave = Arweave.init({
...
-	feeMultiple,
-	arfsDao,
+	dryRun = false,
+	feeMultiple = new FeeMultiple(1.0),
 	appName = DEFAULT_APP_NAME,
 	appVersion = DEFAULT_APP_VERSION,
-	bundle = true
+	walletDao = new WalletDAO(arweave, appName, appVersion),
+	shouldBundle = true,
+	arFSTagBuilder = new ArFSTagSettings({ appName, appVersion }),
```
fedellen, line 54: This var escaped rename

src/ardrive.ts
arielmelendez: ArFSUploadPlanner? I think a lot of what we talk about when prepping uploads is planning.

src/pricing/arfs_cost_estimator.ts
```diff
@@ -89,7 +89,9 @@ export class ArFSCostEstimator {
 		const anchorLength = 1;
 
 		// Get byte length of tags after being serialized for avro schema
-		const serializedTags = serializeTags(this.arFSTagBuilder.withBaseArFSTags(objectPrototype.gqlTags));
+		const serializedTags = serializeTags(
+			this.arFSTagSettings.assembleBaseArFSTags({ tags: objectPrototype.gqlTags })
```
arielmelendez, line 93: This is ok, but there's some "fluent interface" readability opportunities here with the addition of the object params, e.g.:
```
			this.arFSTagSettings.baseTagsIncluding({ tags: objectPrototype.gqlTags })
```
or
```
			this.arFSTagSettings.baseTagsIncluding({ otherTags: objectPrototype.gqlTags })
```

arielmelendez: or:
```
this.arFSTagSettings.baseTags({ includingTags: objectPrototype.gqlTags, excludingTagsNamed: ['foo'] })
```

src/arfs/arfs_tag_settings.test.ts
```diff
@@ -0,0 +1,206 @@
...
+				value: ''
+			};
+
+			expect(() => arFSTagSettings.assertTagLimits([tag])).to.throw(
+				Error,
+				'GQL tag "value" must be a non-empty string!'
+			);
+		});
+
+		it('throws an error if the name field is not a string', () => {
```
arielmelendez, line 194:
```suggestion
		it('throws an error if the value field is not a string', () => {
```

src/arfs/arfs_tag_settings.test.ts
```diff
@@ -0,0 +1,206 @@
...
+				value: 'Name field wrong type test'
+			} as unknown) as GQLTagInterface;
+
+			expect(() => arFSTagSettings.assertTagLimits([tag])).to.throw(
+				Error,
+				'GQL tag "name" must be a non-empty string!'
+			);
+		});
+
+		it('throws an error if the name field is an empty string', () => {
```
arielmelendez, line 182:
```suggestion
		it('throws an error if the value field is an empty string', () => {
```

src/arfs/arfs_tag_settings.test.ts
```diff
@@ -0,0 +1,206 @@
...
+
+	// 	A tag object is valid iff.:
+	// • there are <= 128 tags
+	// • each key is <= 1024 bytes
+	// • each value is <= 3072 bytes
+	// • only contains a key and value
+	// • both the key and value are non-empty strings
+
+	describe('assertTagLimits method', () => {
+		it('throws an error if there are too many GQL tags', () => {
```
arielmelendez, line 108: Please add the negative of this test as well right at the limit, i.e. 128 tags.

src/arfs/arfs_tag_settings.test.ts
```diff
@@ -0,0 +1,206 @@
...
+				extraKey: 'This will error!'
+			} as GQLTagInterface;
+
+			expect(() => arFSTagSettings.assertTagLimits([tag])).to.throw(
+				Error,
+				'GQL tag has too many keys, tags must only have "name" and "value" fields!'
+			);
+		});
+
+		it('throws an error if the name field on a tag has too many bytes', () => {
```
arielmelendez, line 134: Please add the negative of this test as well right at the limit, i.e. 1024 bytes.

src/arfs/arfs_tag_settings.test.ts
```diff
@@ -0,0 +1,206 @@
...
+				value: 'Name field byte limit test'
+			};
+
+			expect(() => arFSTagSettings.assertTagLimits([tag])).to.throw(
+				Error,
+				'GQL tag "name" field byte size (1025) has exceeded the maximum byte limit allowed of 1024!'
+			);
+		});
+
+		it('throws an error if the value field on a tag has too many bytes', () => {
```
arielmelendez, line 146: Please add the negative of this test as well right at the limit, i.e. 3072 bytes.

arielmelendez (changes requested):
Left a handful of test requests, typo fix suggestions, and potential naming update suggestions.

arielmelendez (approved):

PR #112, opened 2021-12-21
PE-821: Incorrect price estimation

fedellen:
This PR replaces the chunk estimator with a new `ARDataPriceNetworkEstimator`. Unfortunately, our assumptions with the arweave pricing algorithm have proven to be false. We are under funding rewards on production.

This new class will cache the pricing results based on how many chunks the byteCount would've been. With chunk based caching we can still be efficient when pricing metaData transactions and base fee transactions. It will get the cached value or simply call the gateway directly if a cached price is not found.

Since this is a hot-fix with the base branch of `master`, here are the proposed GitHub release notes:

- Price estimator now gathers and caches prices directly from the gateway

### DIFF
--- a/src/pricing/ar_data_price_network_estimator.ts
+++ b/src/pricing/ar_data_price_network_estimator.ts
@@ -3,7 +3,7 @@ import type { ArweaveOracle } from './arweave_oracle';
 import { AbstractARDataPriceAndCapacityEstimator } from './ar_data_price_estimator';
 import { AR, ByteCount, Winston, ArDriveCommunityTip } from '../types';
 
-const byteCountOfChunk = new ByteCount(Math.pow(2, 10) * 256); // 256 KiB
+const byteCountPerChunk = new ByteCount(Math.pow(2, 10) * 256); // 256 KiB
 
 /**
  * A utility class for Arweave data pricing estimation.
@@ -11,7 +11,7 @@ const byteCountOfChunk = new ByteCount(Math.pow(2, 10) * 256); // 256 KiB
  */
 export class ARDataPriceNetworkEstimator extends AbstractARDataPriceAndCapacityEstimator {
 	private bytesToChunks(bytes: ByteCount): number {
-		return Math.ceil(+bytes / +byteCountOfChunk);
+		return Math.ceil(+bytes / +byteCountPerChunk);
 	}
 
 	private cachedPricePerChunk: { [chunks: string]: Promise<Winston> } = {};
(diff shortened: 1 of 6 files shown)


src/pricing/ar_data_price_network_estimator.ts
```diff
@@ -0,0 +1,86 @@
+import { GatewayOracle } from './gateway_oracle';
+import type { ArweaveOracle } from './arweave_oracle';
+import { AbstractARDataPriceAndCapacityEstimator } from './ar_data_price_estimator';
+import { AR, ByteCount, Winston, ArDriveCommunityTip } from '../types';
+
+const byteCountOfChunk = new ByteCount(Math.pow(2, 10) * 256); // 256 KiB
```
arielmelendez, line 6:
```suggestion
const byteCountPerChunk = new ByteCount(Math.pow(2, 10) * 256); // 256 KiB
```
You'd need to change it in all usage locations.

src/pricing/ar_data_price_network_estimator.ts
```diff
@@ -0,0 +1,86 @@
...
+	 */
+	public async getBaseWinstonPriceForByteCount(byteCount: ByteCount): Promise<Winston> {
+		const chunks = this.bytesToChunks(byteCount);
+		const cachedPrice = this.cachedPricePerChunk[`${chunks}`];
+
+		if (cachedPrice) {
+			return cachedPrice;
+		}
+
+		const winstonPrice = await this.oracle.getWinstonPriceForByteCount(byteCount);
```
arielmelendez, line 46: This simple implementation should work fine to get us over the hump today, but we should consider a solution like we have in the ArFSCache PR that caches promises so that we BOTH request AND response caching. This solution only has response caching, so you can get cache misses if you issue multiple request for the same chunk size back-to-back while the first one is in flight.

fedellen: I was hoping to use the ArFSCache directly here in the future, if thats possible. But I might be able to adjust this to handle promises

fedellen: cea10bb

src/pricing/ar_data_price_network_estimator.ts
fedellen: 0dd92a8

arielmelendez: Looks like that change got made in the wrong file.

arielmelendez (approved):

PR #113, opened 2021-12-22
PE-829: Price estimators with fallback

fedellen:
This PR adds a fallback estimator that will be utilized in the event of rate limiting by the gateway pricing API. By default, the estimator uses the new network price estimator to ensure our pricing is consistent with the gateway. If the network estimator gets rate limited, it will fallback to the chunk estimator which will use the cached results to calculate rewards.

During the initial pricing request, a new `CachedGatewayOracle` class will first issue requests for base fee and one chunk. This is the information that the chunk estimator needs to calculate price in the event of rate limiting. We also use these two price estimates the most (1 chunk for metadata, 0 chunks for tip transaction reward) so it shouldn't be too wasteful.

The chunk estimator has only been temporarily repaired here. Our understanding of the Arweave pricing algorithm remains uncertain and our previous assumption has been proven false. I've added a new "magicBullet" here that simply adds 1 Winston per chunk. This will cover the pricing as we see it today, but it will need to be monitored as this could increase in the future.

Suggested review order:
- `gateway_oracle.test.ts`
- `gateway_oracle.ts`
- `ar_data_price_fallback_estimator.test.ts`
- `ar_data_price_fallback_estimator.ts`
- `ar_data_price_chunk_estimator.ts`


src/pricing/ar_data_price_fallback_estimator.ts
```diff
@@ -0,0 +1,53 @@
...
+	) {
+		super();
+	}
+
+	private async useClosureAndFallbackOnFail<T>(
+		closure: (priceEstimator: AbstractARDataPriceAndCapacityEstimator) => Promise<T>
+	): Promise<T> {
+		let result: T | undefined = undefined;
+
+		while (!result) {
```
fedellen, line 27: Add retries here

fedellen:
Marking as draft temporarily. Needs updates on upload planner tests as the expected rewards seems to change here, causing the failed test

fedellen:
Fallback may longer be necessary here with rate limiting not being an issue. And using the chunk estimator as a fallback will ultimately be unreliable and accumulate its own tech debt until we know exactly what the pricing algorithm is doing

But having generic retries on the pricing request in case of network timeout and moving the caching to the gateway oracle during this PR will still be useful add here. Probably just remove the inaccurate chunk estimator entirely

Will remain in draft until reprioritized and this PR is updated

fedellen:
De-prioritized and now has conflicts. Closing this for now.

PR #115, opened 2021-12-23
PE-636: Send community tip with the data tx or bundle tx

fedellen:
This PR adds the ArDrive community tip to the bundled transaction or to the data transaction in the case of `--no-bundle`.

Adding support for this was fairly straight forward. The only complication was how to handle the tip parameters on the bulk folder upload method. Given that my intent is to deprecate the bulk folder method and create a new `uploadAllEntities` method in the bulk bundle branch, I have temporarily made `uploadAsV2Tx` accessible and have used it directly to avoid the conflict.

The unnecessary tip tag `"Type"` has been removed during this PR in compliance with the [GQL Tag Usage Doc](https://ardrive.atlassian.net/wiki/spaces/ENGINEERIN/pages/194969618/Graph+Query+Language+Tags+and+Usage#Community-Tip-Transactions)


src/arfs/arfsdao.ts
```diff
@@ -755,6 +787,15 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 			txAttributes.last_tx = 'STUB';
 		}
 
+		if (communityTipSettings) {
+			// Add community tip to the bundle trx
```
fedellen, line 791:
```suggestion
			// Add community tip to the v2 transaction
```

matibat (commented):
It's looking good. Gotta review the CLI side

matibat (approved):
Looks good!

jdaev (approved):
Looks good

src/ardrive.ts
```diff
@@ -483,10 +491,12 @@ export class ArDrive extends ArDriveAnonymous {
 					key: isPrivateResult(uploadFileResult) ? urlEncodeHashKey(uploadFileResult.fileKey) : undefined
 				}
 			],
-			tips: [tipData],
+			tips: [],
 			fees: {}
 		};
 
+		const tipResult = { recipient: communityTipTarget, winston: communityWinstonTip };
```
arielmelendez, line 498: Consider typing this as `Partial<TipData>` to indicate to the reader that it's an incomplete, typed object.

fedellen: Went with `Omit<TipData, 'txId'>` to keep the compiler happy, it didn't like the optionals: d09e1e0

src/arfs/arfsdao.ts
```diff
@@ -635,16 +636,20 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 		return { arFSObjects: [dataArFSObject, metaDataArFSObject], fileId };
 	}
 
-	private async uploadFileV2Tx(
+	// Temporarily public for bulk folder use case, which will be deprecated in PE-460
+	// Re-make this private once `uploadAllEntities` exists
+	/* private */ async uploadFileV2Tx(
```
arielmelendez, line 641: Consider marking as deprecated immediately

arielmelendez (approved):

PR #120, opened 2022-01-12
PE-460: Bulk upload as bundles

fedellen:
This PR adds the `uploadAllEntities` flow for all cases of uploading files. The old single/bulk methods have been deprecated and now use the `uploadAllEntities` internally. The integration tests for the single file flow have been preserved to ensure that the API remains backwards compatible.

During this PR, the `UploadPlanner` class has been heavily modified and is focused only on determining the byteCounts of the upload and planning the upload strategy. A new class, `CostCalculator`, consumes the byteCounts from the upload planner and determines the proper `rewardSettings` and `communityTipSettings` for each bundle or v2 transaction plan.

There is a heavy emphasis on unit/integration testing here because the previous bulk methods were mostly untested. Now we have coverage on all of the bulk methods.

Suggested review order:

`ardrive.ts`
`arfs_upload_planner.test.ts`
`arfs_upload_planner.ts`
`arfs_cost_calculator.test.ts`
`arfs_cost_calculator.ts`
`arfsdao.test.ts`
`arfsdao.ts`
`ardrive.int.test.ts`


tests/integration/ardrive.int.test.ts
```diff
@@ -615,18 +625,21 @@ describe('ArDrive class - integrated', () => {
 					assertUploadFileExpectations(result, W(3204), W(171), W(1), 'public', existingFileId);
 				});
 
-				it('throws an error if destination folder has a conflicting FILE name and a matching last modified date and the conflict resolution is set to upsert', async () => {
+				it('returns an empty ArFSResult if destination folder has a conflicting FILE name and a matching last modified date and the conflict resolution is set to upsert', async () => {
```
fedellen, line 628: Note for the reviewer:

We return an empty result in this case now so we can continue with other parts of the upload. Previously in the single file only use case, we returned an error explaining why the file was not uploaded.

tests/integration/ardrive.int.test.ts
```diff
@@ -1172,6 +1189,226 @@ describe('ArDrive class - integrated', () => {
 			});
 		});
 	});
+
+	const stubNameConflictInfo = {
```
fedellen, line 1193: Re-use this info in existing tests

src/arfs/arfsdao.ts
```diff
@@ -637,15 +574,112 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
+	async uploadPublicFile({
+		parentFolderId: destFolderId,
+		wrappedFile: wrappedEntity,
+		driveId: destDriveId,
+		rewardSettings,
+		communityTipSettings
+	}: ArFSUploadPublicFileParams): Promise<ArFSUploadEntitiesResult> {
+		const uploadPlan: CalculatedUploadPlan = ((): CalculatedUploadPlan => {
+			if (isBundleRewardSetting(rewardSettings)) {
+				rewardSettings.bundleRewardSettings;
```
fedellen, line 596:
```suggestion
```

src/arfs/arfsdao.ts
```diff
@@ -749,55 +798,186 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
 						fileId,
-						parentFolderId
+						destFolderId
 					)
 				)
 		};
+	}
 
-		return this.uploadFile(prepFileParams, rewardSettings, communityTipSettings);
+	/** Assembles folder metadata prototype factory to be used in prepareFile */
```
fedellen, line 809:
```suggestion
	/** Assembles folder metadata prototype factory to be used in prepareFolder */
```

src/arfs/arfsdao.ts
```diff
@@ -660,80 +692,97 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
-
-	async uploadPublicFile({
-		parentFolderId,
-		wrappedFile,
-		driveId,
-		rewardSettings,
-		communityTipSettings
-	}: ArFSUploadPublicFileParams): Promise<ArFSUploadFileResult> {
+	/** Assembles data and metadata prototype factories to be used in prepareFile */
+	private getPrepFileParams({
```
fedellen, line 749: These convenience factory assembly functions don't refer to `this` anywhere. Perhaps they don't belong in the ArFSDAO class

src/ardrive.ts
```diff
@@ -419,342 +391,220 @@ export class ArDrive extends ArDriveAnonymous {
...
+		for (const { destFolderId, destName, wrappedEntity, driveKey } of entitiesToUpload) {
+			const destDriveId = await this.arFsDao.getDriveIdForFolderId(destFolderId);
+			// Assert drive privacy and owner of the drive
+			const owner = await this.arFsDao.getOwnerAndAssertDrive(destDriveId, driveKey);
+			await this.assertOwnerAddress(owner);
+
+			// Assign destination name and derive names already within provided destination folder
+			wrappedEntity.newName = destName;
+			const nameConflictInfo = driveKey
+				? await this.arFsDao.getPrivateNameConflictInfoInFolder(destFolderId, driveKey)
```
arielmelendez, line 415: Having hit that scalability issue in that user's drive last night, I'd say that this is going to be a non-starter without some intelligent caching and/or a change in the algorithm here. Cache management will be challenging here because:
• We fetch the folder listing from gql and assemble entities (expensive on big drives)
• We COULD, but currently don't, cache that list.
• We confirm that there's no conflict and go ahead with the upload. This should now cause a cache invalidation. UNLESS...
• We dynamically update our cache entry. BUT in order to do so, we need more info than we have on hand to create the entity to do so (i.e. the file entity that would go in the cached entity listing for the folder); the data we SEND to the network is different than the data we retrieve. If we had a way to map from the former to the latter we could perhaps manage this.

Another way we could do this right now is:
- Reduce out map whose keys are a tuple key type of folderID/driveKey and whose values are associated entities from the entitiesToUpload list.
- For each key in that map, one at a time:
- request the entity list for the folder and then:
- Check for naming conflicts against each of the files to upload associated with the key (also make sure there's no conflicts between the files to upload themselves! i.e. myfile.png from source folder A and myfile.png from source folder B)
- If everything is ok, proceed with the existing loop in your code.

arielmelendez (changes requested):
Identified a major scaling issue.

src/ardrive.ts
arielmelendez: > the data we SEND to the network is different than the data we retrieve.
Just thought about this again... We need to cache (and update) the list of CONFLICT INFOS for which I believe we'd now have enough info. So if we had a folder listing cache (which we could safely invalidate) and a conflict info cache (which we could safely update on the fly) we might be in good shape.

fedellen: I have addressed this by adding conflict info caching to the existing in memory `ArFSEntityCache` and also asserting any local conflicts in the upload itself before resolving the on-chain data. Each subsequent call to the ArFSDAO for the same conflict info for a given destination folder will return the existing conflicts info from the cache.

Because all of the conflict resolution happens before the uploading process, there is no reason to dynamically update or to invalidate the cache as of yet

src/arfs/arfsdao.ts
```diff
@@ -241,6 +172,7 @@ export interface ArFSCache extends ArFSAnonymousCache {
 	privateDriveCache: ArFSEntityCache<ArFSPrivateDriveCacheKey, ArFSPrivateDrive>;
 	privateFolderCache: ArFSEntityCache<ArFSPrivateFolderCacheKey, ArFSPrivateFolder>;
 	privateFileCache: ArFSEntityCache<ArFSPrivateFileCacheKey, ArFSPrivateFile>;
+	privateConflictCache: ArFSEntityCache<ArFSPrivateFolderCacheKey, NameConflictInfo>;
```
arielmelendez, line 175: Consider moving this to the ArFSCache class as it's never used in the anonymous context.

src/utils/upload_conflict_resolution.ts
```diff
@@ -6,20 +7,70 @@ import {
...
+		if (existingName) {
+			throw new Error('Upload cannot contain multiple destination names to the same destination folder!');
+		}
+
+		if (wrappedEntity.entityType === 'folder') {
+			assertConflictsWithinFolder(wrappedEntity);
+		}
+
+		// Add local upload info to check for name conflicts within the upload itself
+		namesWithinUpload[`${destFolderId}`] = namesWithinUpload[`${destFolderId}`]
```
arielmelendez, line 33: That spread might get unnecessarily wasteful for large upload lists. Consider:
```
		if (!namesWithinUpload[`${destFolderId}`]) {
			namesWithinUpload[`${destFolderId}`] = [];
		}
		namesWithinUpload[`${destFolderId}`].push(destinationName);
```

src/utils/upload_conflict_resolution.ts
```diff
@@ -6,20 +7,70 @@ import {
...
+	for (const { destFolderId, wrappedEntity, destName } of entitiesToUpload) {
+		const destinationName = destName ?? wrappedEntity.destinationBaseName;
+
+		const existingName = namesWithinUpload[`${destFolderId}`]?.find((n) => n === destinationName);
+		if (existingName) {
+			throw new Error('Upload cannot contain multiple destination names to the same destination folder!');
+		}
+
+		if (wrappedEntity.entityType === 'folder') {
+			assertConflictsWithinFolder(wrappedEntity);
```
arielmelendez, line 29: I believe for the logic of the overarching function to be correct, `assertConflictsWithinFolder` needs to bear a reference to `namesWithinUpload` recursively with it. Otherwise, I'd be surprised if there were ever a conflict arising from that function since the host file system has probably been doing a pretty good job of keeping folders free from duplicate names. That's not to say that a caller couldn't pass any mangled tree of wrapped entities they wanted through though...

Make sure your tests are adequately covering this case. (I'll check too, and I may have misinterpreted what's taking place in the recursive function).

arielmelendez: Welp, there appears to be no unit test coverage of these functions specifically. Your integration tests seem to be missing the case where a file in a nested folder has the same name as a file in a folder above or below it. I've just skimmed quickly, so let me know if I've missed something.

arielmelendez (changes requested):
I've only had time to review the conflict info caching so far, but I think it's got a bug as per my comments.

src/utils/upload_conflict_resolution.ts
fedellen: But there would be no conflict if a file/folder had the same name as another file/folder above or below it. They are being sent to different destination folders

```
// This is fine, files are being uploaded to different destinations
file-to-upload.txt 
folder-to-upload
     file-to-upload.txt
     
// This will error
folder-to-upload
     file-to-upload.txt
     file-to-upload.txt 
```

`assertConflictsWithinFolder` does not need the `namesWithinUpload` information because each folder in the upload plan will be a unique entity and the entities inside of it are locked into the recursive structure, they cannot have a destination folder other than the folder they reside in. We also don't have folderId information for the folderToUpload at this point -- although it is an option to to refactor that up from the uploadPlanner layer and instead perform the flattening at the ArDrive layer before conflict resolution.

src/arfs/arfsdao.ts
fedellen: This class?

```ts
export class ArFSEntityCache<K, V> {
	private cache: Cache<string, Promise<V>>;

	constructor(capacity: number) {
		this.cache = SimpleCache<string, Promise<V>>(capacity);
	}
```

Or do you mean remove the `arfsdao_anonymous.ts` side of the cache and put bring them both to the writing class?

src/utils/upload_conflict_resolution.ts
fedellen: Added: 6ec03ab

src/utils/upload_conflict_resolution.test.ts
```diff
@@ -417,11 +445,125 @@ describe('The resolveFolderNameConflicts function', () => {
...
+	});
+
+	it('succeeds without error if an upload has a multiple files with the same destName but different destFolderIds', () => {
+		const file1UploadStats: UploadStats = { ...stubFileUploadStats(stubEntityID), destName: 'same-name' };
+		const file2UploadStats: UploadStats = { ...stubFileUploadStats(stubEntityIDAlt), destName: 'same-name' };
+
+		expect(() => assertLocalNameConflicts([file1UploadStats, file2UploadStats])).to.not.throw(Error);
+	});
+
+	it('succeeds without error if an upload has a multiple folders with the same destFolderId but different destNames', () => {
```
fedellen, line 563:
```suggestion
	it('succeeds without error if an upload has multiple files with the same destFolderId but different destNames', () => {
```

fedellen: Multiple test cases here need rephrasing

src/arfs/arfsdao.ts
matibat: I'd say the latter makes more sense

matibat (commented):
Very robust test set! I like the clever approach of using the cache on the conflicts and the usage of the factory functions to generalize the algorithm regardless of the drive privacy.

src/arfs/arfs_cost_calculator.ts
```diff
@@ -0,0 +1,202 @@
...
+	}: V2TxPlan): Promise<{ calculatedV2TxPlan: CalculatedV2TxPlan; totalPriceOfV2Tx: Winston }> {
+		let totalPriceOfV2Tx: Winston = W(0);
+		let rewardSettings: Partial<UploadFileV2TxRewardSettings> = {};
+		let communityTipSettings: CommunityTipSettings | undefined = undefined;
+
+		if (fileDataByteCount) {
+			const winstonPriceOfDataTx = await this.priceEstimator.getBaseWinstonPriceForByteCount(fileDataByteCount);
+			totalPriceOfV2Tx = totalPriceOfV2Tx.plus(this.boostedReward(winstonPriceOfDataTx));
+
+			rewardSettings = {
```
arielmelendez, line 102: No need to spread here or on 122.

src/pricing/estimation_prototypes.ts
```diff
@@ -11,49 +16,69 @@ import {
...
+	readJWKFile
 } from '../exports';
 import { EstimateCreateDriveParams } from '../types/upload_planner_types';
 import { fakeEntityId, fakeTxID } from '../utils/constants';
 
+export const getFakeDriveKey = async (): Promise<DriveKey> => {
+	return deriveDriveKey(
+		'stubPassword',
+		`${fakeEntityId}`,
+		JSON.stringify((readJWKFile('./test_wallet.json') as JWKWallet).getPrivateKey())
```
arielmelendez, line 35: We shouldn't be shipping this file. Let's just put a hard coded fake key in.

src/pricing/estimation_prototypes.ts
```diff
@@ -11,49 +16,69 @@ import {
...
+	);
+};
+
+export async function getPrivateFolderEstimationPrototype(
```
arielmelendez, line 39: Could use docs on all the functions below.

src/utils/bundle_packer.ts
```diff
@@ -0,0 +1,85 @@
...
+
+interface BundlePackParams {
+	// Upload stats can be undefined when a metaData for an over-sized file is packed
+	uploadStats?: UploadStats;
+	byteCountAsDataItem: ByteCount;
+	numberOfDataItems: number;
+}
+
+export abstract class BundlePacker {
+	public bundles: BundleToPack[] = [];
```
arielmelendez, line 13: Can we add a getter and make this protected so that the reference can't be modified by clients?

src/utils/bundle_packer.ts
```diff
@@ -0,0 +1,85 @@
...
+		// Otherwise we pack into a new bundle
+		this.bundles.push(new BundleToPack(bundlePackParams, this.maxBundleSize, this.maxDataItemLimit));
+		return this.bundles.length - 1;
+	}
+}
+
+class BundleToPack {
+	public uploadStats: UploadStats[] = [];
+
+	public totalSize = 0;
```
arielmelendez, line 58: These should not be public. Make getters. And if totalDataItems could be a getter that returns the sum of 2 lists (e.g. uploadStats's count and another list, like metadatas?, count) that might clean some of this up.

src/arfs/arfs_upload_planner.ts
```diff
@@ -1,168 +1,291 @@
...
+			} else {
+				this.v2TxsToUpload.push({
+					uploadStats: planFolderParams,
+					metaDataByteCount: folderMetaDataPrototype.objectData.sizeOf()
+				});
+			}
+		}
+		// For new folder creations, we will generate and preserve the folder ID early here to prevent
+		// the parent to child folder relationship from being lost during the flattening of the folder tree
+		wrappedFolder.existingId ??= EID(v4());
```
arielmelendez, line 157: Consider moving this into the conditional above without the `??` and maybe update the comment to indicate that "folder IDs must be established beyond this point, so generate new ones for any folders that don't appear to exist on chain yet".

src/arfs/arfs_upload_planner.ts
```diff
@@ -1,168 +1,291 @@
...
+				await this.planFile({ ...uploadStat, wrappedEntity, isBulkUpload });
+			}
 		}
 
-		return this.costOfCreateDriveV2Tx(arFSPrototypes);
-	}
+		const bundlePlans: BundlePlan[] = [];
+		for (const { uploadStats, totalDataItems, totalSize } of this.bundlePacker.bundles) {
+			if (totalDataItems === 1) {
+				// Edge case: Do not send up a bundle with a single folder data item
```
arielmelendez, line 223: How do we reliably know that this means there's a folder involved?

src/arfs/arfs_upload_planner.ts
```diff
@@ -193,32 +316,14 @@ export class ArFSUploadPlanner {
 		return new ByteCount(totalByteLength);
 	}
 
-	/** Calculate the bundled size from an array of data item byte counts  */
-	private bundledByteCountOfDataItems(dataItemByteCounts: ByteCount[]): ByteCount {
+	/** Calculate the bundled size from the total dataItem byteCount and the number of dataItems */
+	private bundledByteCountOfBundleToPack(totalDataItemByteCount: ByteCount, numberOfDataItems: number): ByteCount {
```
arielmelendez, line 320: Does this need to be a private member function?

src/utils/bundle_packer.ts
fedellen: I added these as protected members with getters here: 4223eb2

I tried out a larger refactor that removed the totalDataItems but ran into more permutations such as needing to know whether a file UploadStat will be bundled with its metadata or not. I think I'd need to generalize v2/bundle and handle it all here as PlannedUploads. I'll circle around back to this try it out with that approach later on

src/utils/bundle_packer.ts
```diff
@@ -0,0 +1,85 @@
...
+export abstract class BundlePacker {
+	public bundles: BundleToPack[] = [];
+
+	abstract packIntoBundle(bundlePackParams: BundlePackParams): BundleIndex;
+}
+
+/**
+ * Pack into lowest index bundle with available size and remaining data items
+ *
+ * Preserve the BundleIndex for use in edge case where FileData is above MAX_BUNDLE_SIZE
```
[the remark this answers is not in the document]
fedellen: Included here: 4223eb2

src/utils/bundle_packer.ts
fedellen: Included here: 4223eb2

src/pricing/estimation_prototypes.ts
fedellen: Added, and re-ordered the functions: 403b139

src/pricing/estimation_prototypes.ts
fedellen: I've changed it to use stub string: 25ed5e7

I tried moving this file to the stubs folder so we remove it from the produced source but have run into lots of hardcoded paths and stashed that for now. Will try to solve this later

src/arfs/arfs_cost_calculator.ts
fedellen: Fixed: 4784ead

src/arfs/arfs_upload_planner.ts
fedellen: 5bccf9a

src/arfs/arfs_upload_planner.ts
fedellen: Made them pure functions for now: 2e18982

src/arfs/arfs_upload_planner.ts
fedellen: Added more info here: 513cf68

src/utils/bundle_packer.ts
```diff
@@ -0,0 +1,97 @@
...
+
+interface BundlePackParams {
+	// Upload stats can be undefined when a metaData for an over-sized file is packed
+	uploadStats?: UploadStats;
+	byteCountAsDataItem: ByteCount;
+	numberOfDataItems: number;
+}
+
+export abstract class BundlePacker {
+	constructor(protected readonly maxBundleSize: ByteCount, protected readonly maxDataItemLimit: number) {}
```
arielmelendez, line 13: Nit: throw if maxDataItemLimit is not a finite positive integer.

src/arfs/arfs_upload_planner.ts
```diff
@@ -1,224 +1,340 @@
...
+		maxDataItemLimit = MAX_DATA_ITEM_LIMIT,
+		bundlePacker = (maxBundleSize, maxDataItems) => new LowestIndexBundlePacker(maxBundleSize, maxDataItems)
 	}: ArFSUploadPlannerConstructorParams) {
-		this.priceEstimator = priceEstimator;
 		this.shouldBundle = shouldBundle;
-		this.feeMultiple = feeMultiple;
 		this.arFSTagSettings = arFSTagSettings;
-		this.communityOracle = communityOracle;
-	}
+		this.maxBundleLimit = maxBundleLimit;
```
arielmelendez, line 52: Nit: throw if maxDataItemLimit is not a finite positive integer.

src/types/upload_planner_types.ts
```diff
@@ -1,19 +1,27 @@
...
 } from '../arfs/arfs_prototypes';
 import { ArFSTagSettings } from '../arfs/arfs_tag_settings';
 import { CommunityOracle } from '../community/community_oracle';
+import { FileUploadStats, FolderUploadStats, UploadStats } from '../exports';
 import { ARDataPriceEstimator } from '../pricing/ar_data_price_estimator';
+import { BundleIndex, BundlePacker } from '../utils/bundle_packer';
 import { ByteCount } from './byte_count';
 import { GQLTagInterface } from './gql_Types';
 
+export type BundlePackerFactory = (maxBundleSize: ByteCount, maxDataItems: number) => BundlePacker;
```
arielmelendez, line 16: I think this approach is somewhat limiting because it presumes that every BundlePacker instance would be constructed with these inputs. Since we have the luxury of not needing special and/or transient computed values to be passed to us as parameters by the UploadPlanner, we can simply remove the inputs here and rely upon closure-captured inputs to feed the code inside the factory function. Since we're only using it in a small handful of places I think we'll be ok e.g.:
```
	constructor({
		shouldBundle = true,
		arFSTagSettings,
		maxBundleLimit = MAX_BUNDLE_SIZE,
		maxDataItemLimit = MAX_DATA_ITEM_LIMIT,
		bundlePacker = () => new LowestIndexBundlePacker(MAX_BUNDLE_SIZE, MAX_DATA_ITEM_LIMIT)
	}: ArFSUploadPlannerConstructorParams) {
```

arielmelendez: You can actually take this further by removing maxBundleLimit and maxDataItemLimit from the upload planner and adding an API to bundle packer like:
```
public abstract canPackDataItemsWithSizes(sizes: ByteCount[]): boolean;
```
You'd be able to validate that you're not putting too many data items AND/OR too much data into the bundle via an API like that and you could use it in the uploadplanner code like so:
```
		const totalByteCountOfFileDataItems = new ByteCount(+fileDataItemByteCount + +metaDataByteCountAsDataItem);

		if (
			!this.shouldBundle ||
			!bundlePacker.canPackDataItemsWithSizes([fileDataItemByteCount, metaDataByteCountAsDataItem])
		) {
			// If the file data is too large it must be sent as a v2 tx
			const v2TxToUpload: V2TxPlan = { uploadStats: planFileParams, fileDataByteCount };
```

arielmelendez: The logic you'd need for the function above looks like it's implemented here:
```
	public packIntoBundle(bundlePackParams: BundlePackParams): BundleIndex {
		const { byteCountAsDataItem, numberOfDataItems } = bundlePackParams;

		if (+byteCountAsDataItem > +this.maxBundleSize) {
			throw new Error('Data item is too large to be packed into a bundle!');
		}

		if (numberOfDataItems > this.maxDataItemLimit) {
			throw new Error('Data item count is too high to be packed into a bundle!');
		}
```

src/utils/bundle_packer.ts
```diff
@@ -0,0 +1,97 @@
+import { ByteCount, UploadStats } from '../types';
+
+export type BundleIndex = number;
+
+interface BundlePackParams {
+	// Upload stats can be undefined when a metaData for an over-sized file is packed
+	uploadStats?: UploadStats;
+	byteCountAsDataItem: ByteCount;
```
arielmelendez, line 8: Nit: name would be more generalized as `byteCountAsDataItems`

src/arfs/arfs_upload_planner.ts
```diff
@@ -1,224 +1,340 @@
...
+				}
+			}
 		}
 
-		return this.costOfCreateDriveV2Tx(arFSPrototypes);
-	}
+		const bundlePlans: BundlePlan[] = [];
+		for (const { uploadStats, totalDataItems, totalSize } of bundlePacker.bundles) {
+			if (totalDataItems === 1) {
+				// Edge case: Do not send up a bundle with a data item
```
arielmelendez, line 237:
```suggestion
				// Edge case: Do not send up a bundle with a single data item
```

src/arfs/arfsdao.ts
```diff
@@ -668,144 +706,203 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
+					// Send file as v2, but prepare the metadata as data item
+					const { fileResult, metaDataDataItem } = await this.uploadOnlyFileAsV2(
+						prepFileParams,
+						dataTxRewardSettings,
+						communityTipSettings
+					);
+					v2FileResult = fileResult;
+
+					// Add data item to its intended bundle plan
+					bundlePlans[metaDataBundleIndex].metaDataDataItems === undefined
```
arielmelendez, line 804: Is this just the case when we have a handful of V2 txs whose metadatas can get bundled together? e.g. I have 2 files that are too large for bundling but we can send their 2 metadata txs together in a bundle? It just seems surprising that there would be a bundle with an undefined (rather than empty) metaDataItems.

arielmelendez: It's a little weird to use a ternary without collecting the resulting value. I'd suggest updating to:
```
					const bundlePlan = bundlePlans[metaDataBundleIndex];
					if (bundlePlan.metaDataDataItems === undefined) {
						bundlePlan.metaDataDataItems = [metaDataDataItem];
					} else {
						bundlePlan.metaDataDataItems.push(metaDataDataItem);
					}
```
This approach also allows you to remove the `?.` since the type checker has verified the existence of the array.

src/arfs/arfsdao.ts
```diff
@@ -668,144 +706,203 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
 
-				// Preserve file key on private metadata for return
-				fileKey = metaDataTxData.fileKey;
+				results.folderResults.push({
+					folderId,
+					folderTxId: metaDataTxId,
+					folderMetaDataReward: metaDataTxReward,
+					driveKey
+				});
+			}
```
arielmelendez, line 826: No else case... Is it an error if neither dataTxRewardSettings nor metaDataTxRewardSettings are defined?

src/utils/bundle_packer.ts
```diff
@@ -0,0 +1,97 @@
+import { ByteCount, UploadStats } from '../types';
+
+export type BundleIndex = number;
+
+interface BundlePackParams {
+	// Upload stats can be undefined when a metaData for an over-sized file is packed
+	uploadStats?: UploadStats;
```
arielmelendez, line 7: These optional upload stats still give me "the willies". I wish we had an object oriented way to deal with it. Also, are BundlePackParams really just a `DataItemPlan`?

src/utils/bundle_packer.ts
```diff
@@ -0,0 +1,97 @@
...
+	byteCountAsDataItem: ByteCount;
+	numberOfDataItems: number;
+}
+
+export abstract class BundlePacker {
+	constructor(protected readonly maxBundleSize: ByteCount, protected readonly maxDataItemLimit: number) {}
+
+	protected plannedBundles: PlannedBundle[] = [];
```
arielmelendez, line 15: Is it strictly necessary to provide memory for these at this level? Seems like all that matters is you can get a list out (the getter below). The memory for the list can be discretionary for the packer. NBD though.

src/types/upload_planner_types.ts
fedellen: Applied this here: 91f6b9c

src/arfs/arfsdao.ts
fedellen: Added an error for this case here: 76668e2

Thinking about changing v2TxPlans to instead hold 3 types of plans inside it -- `FileDataOnlyPlan[]`, `FileWithMetaDataPlan[]`, and `FolderMetaDataPlan[]`. These plans would have their shapes defined and then we could handle this with types instead of checking for invalid and throwing errors

src/arfs/arfsdao.ts
fedellen: Updated with just using an empty array here to reduce the conditionals as in your first suggestion: 16f52b5

src/utils/bundle_packer.ts
fedellen: 22acb1b

src/utils/bundle_packer.ts
```diff
@@ -0,0 +1,107 @@
...
+}
+
+export abstract class BundlePacker {
+	constructor(protected readonly maxBundleSize: ByteCount, protected readonly maxDataItemLimit: number) {
+		if (!Number.isFinite(maxDataItemLimit) || !Number.isInteger(maxDataItemLimit) || maxDataItemLimit < 2) {
+			throw new Error('Maximum data item limit must be an integer value of 2 or more!');
+		}
+	}
+
+	public abstract bundles: PlannedBundle[];
```
arielmelendez, line 19:
```suggestion
	public abstract readonly bundles: PlannedBundle[];
```
`readonly` will give us most of what we were actually looking for when I suggested leaving memory allocation to the subtypes and providing a getter. And in fact, it works fine with LowestIndexBundlePacker having the protected member and the public getter as you've already done. There should't be any reason for a client of this class to change the reference to the array externally.

src/arfs/arfsdao.ts
arielmelendez: Yea I can't help but shake the idea that the type design here is not quite where we'd like it. Checking for the presence of a reward settings to decide what logic to run seems like we missed an opportunity upstream to use types to define the work that needs to be done here. Your suggestion gets us much closer, if not exactly where we want to be, i.e. simply iterating over list of objects that have exactly what you need in them and that require no guesswork or optional checking to accomplish the uploads.

src/arfs/arfsdao.ts
```diff
@@ -668,144 +708,201 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
 			}
-		};
 
-		const uploadFileResult = await this.uploadFile(prepFileParams, rewardSettings, communityTipSettings);
+			// Add any metaData data items from over-sized files sent as v2
+			dataItems.push(...metaDataDataItems);
 
-		// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-		return { ...uploadFileResult, fileKey: fileKey! };
+			const bundle = await this.prepareArFSObjectBundle({
```
arielmelendez, line 886: Nit: consider `bundleTx` over `bundle`

src/arfs/arfsdao.ts
```diff
@@ -1092,6 +1190,7 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 
 	async getEntitiesInFolder(
 		parentFolderId: FolderID,
+		owner: ArweaveAddress,
```
arielmelendez, line 1193: Why all the changes here and below to require passing the owner through? Why not use a built in to retrieve it based on the FolderID? Seems like this makes it harder to use this function correctly.

src/pricing/estimation_prototypes.ts
```diff
@@ -11,49 +16,88 @@ import {
...
-			await ArFSPrivateFolderTransactionData.from(driveName, newPrivateDriveData.driveKey)
-		),
-		driveMetaDataPrototype: new ArFSPrivateDriveMetaDataPrototype(
-			fakeEntityId,
-			await ArFSPrivateDriveTransactionData.from(driveName, fakeEntityId, newPrivateDriveData.driveKey)
-		)
-	};
+/** Derive a fake drive key from a stub drive key string for estimation and upload planning purposes */
+export const getFakeDriveKey = async (): Promise<DriveKey> => {
+	const fakeDriveKeyString = 'VTAOuxuewZZZZZZZZZZZZHipwJKXzXKxvZaKqFAKE/s';
```
arielmelendez, line 30: Consider making this string even more fake. It looks like it came from a real drive key and now only requires about 4 characters to crack and some low grade brute forcing to figure out what drive it belongs to.

src/types/byte_count.ts
```diff
@@ -15,6 +15,14 @@ export class ByteCount implements Equatable<ByteCount> {
 		return this.byteCount;
 	}
 
+	plus(byteCount: ByteCount): ByteCount {
```
arielmelendez, line 18: These need unit tests

arielmelendez (changes requested):
A few small issues (need a few unit tests, variable names, etc.) and some larger ones that don't necessary need to be tackled right now (e.g. uploadAllEntities-related changes).

src/arfs/arfsdao.ts
fedellen: This was to remain consistent with the other behaviors of the ArFSDAO getters e.g. `getPrivateFile`, `getPublicDrive`, `listPrivateFolder`. If we derive the owner from the folder ID in the ArFSDAO layer, we are not allowing for any flow that would use a custom owner

src/arfs/arfsdao.ts
fedellen: Applied this refactor here: fb4680f

Removes quite a few error cases for invalid plans

src/utils/bundle_packer.ts
fedellen: 3dba5ea

src/arfs/arfsdao.ts
fedellen: da4e543

src/pricing/estimation_prototypes.ts
fedellen: 07c40dd

src/types/byte_count.ts
fedellen: Added: f319a2a

src/utils/bundle_packer.ts
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	public packIntoBundle(dataItemPlan: DataItemPlan): BundleIndex {
+		const { byteCountAsDataItems: byteCountAsDataItem, numberOfDataItems } = dataItemPlan;
+
+		for (let index = 0; index < this.bundles.length; index++) {
+			const bundle = this.bundles[index];
+			// Pack into lowest index bundle that has enough remaining size and data items
+			if (
+				bundle.remainingSize.isGreaterThan(byteCountAsDataItem) &&
```
arielmelendez, line 60: Should be `isGreaterThanOrEqualTo`???

arielmelendez: If so, add a test to catch this case.

arielmelendez: I modified one of your existing test cases to demonstrate the issue:
```
it('packs an entity that would exceed the max size of the first bundle into a second bundle', () => {
			bundlePacker.packIntoBundle({
				uploadStats: fileUploadStats,
				byteCountAsDataItems: new ByteCount(50),
				numberOfDataItems: 2
			});

			expect(bundlePacker.bundles.length).to.equal(1);
			expect(+bundlePacker.bundles[0].totalSize).to.equal(50);
			expect(bundlePacker.bundles[0].totalDataItems).to.equal(2);

			bundlePacker.packIntoBundle({
				uploadStats: fileUploadStats,
				byteCountAsDataItems: new ByteCount(50),
				numberOfDataItems: 2
			});

			expect(bundlePacker.bundles.length).to.equal(1);
			expect(+bundlePacker.bundles[0].totalSize).to.equal(100);
			expect(bundlePacker.bundles[0].totalDataItems).to.equal(2);
		});
```

arielmelendez (changes requested):
Fix boundary case issue and add test to catch it.

src/utils/bundle_packer.ts
fedellen: Oops, good catch there!

Added new byte count method and tests: fe74420
Regression test for the border case: cbabed6
And the fix: 9131134

arielmelendez (approved):

src/utils/upload_conflict_resolution.ts
```diff
@@ -276,11 +277,41 @@ export async function resolveFolderNameConflicts({
...
 				prompts
 			});
 		}
+
+		assertAndRemoveConflictingEntities(wrappedFolder);
+	}
+}
+
+/** Uses conflictResolution on each file and folder to recursively remove skipped entities or error on conflicts */
+export function assertAndRemoveConflictingEntities(folder: ArFSFolderToUpload): void {
```
arielmelendez, line 291: Nit: Pure input/output might be preferable here to mutating the input array.

src/utils/upload_conflict_resolution.ts
```diff
@@ -276,11 +277,41 @@ export async function resolveFolderNameConflicts({
...
+	for (let index = 0; index < folder.files.length; index++) {
+		const childFile = folder.files[index];
+
+		if (childFile.conflictResolution === 'skip' || childFile.conflictResolution === 'upsert') {
+			// Remove from intended files if file will be skipped
+			folder.files.splice(index, 1);
```
arielmelendez, line 297: I think there's a subtle bug here. You shouldn't iterate the loop if you remove this item, because the next item in the array is going to enter the current index and you want to check that one. Remove your post loop step and iterate manually. ALTERNATIVELY, and more safely in my opinion, start from the end of the array and iterate backwards in the post loop step.

arielmelendez (changes requested):
Fix subtle looping bug

src/utils/upload_conflict_resolution.ts
arielmelendez: This works for me:
```
/** Uses conflictResolution on each file and folder to recursively remove skipped entities or error on conflicts */
export function assertAndRemoveConflictingEntities(folder: ArFSFolderToUpload): void {
	let index = folder.files.length;
	while (index--) {
		const childFile = folder.files[index];

		if (childFile.conflictResolution === 'skip' || childFile.conflictResolution === 'upsert') {
			// Remove from intended files if file will be skipped
			folder.files.splice(index, 1);
		} else if (childFile.conflictResolution === 'error') {
			throw Error(errorMessage.entityNameExists);
		}
	}

	index = folder.folders.length;
	while (index--) {
		const childFolder = folder.folders[index];

		if (childFolder.conflictResolution === 'skip') {
			// Remove from intended folders if folder will be skipped
			folder.folders.splice(index, 1);
		} else if (childFolder.conflictResolution === 'error') {
			throw Error(errorMessage.entityNameExists);
		} else {
			// Recurse into folder
			assertAndRemoveConflictingEntities(childFolder);
		}
	}
}
```

tests/integration/ardrive.int.test.ts
```diff
@@ -1557,6 +1564,77 @@ describe('ArDrive class - integrated', () => {
 			assertTipSetting(tips[0], bundleTxId!);
 		});
 
+		it('returns the expected empty ArFSResult with a folder that has multiple conflicting names within its tree and uses --skip conflictResolution', async () => {
```
arielmelendez, line 1567: Nice

arielmelendez (approved):

src/arfs/arfsdao.ts
matibat: Agreed on the latter

src/utils/upload_conflict_resolution.ts
matibat: I assume this suggestion was applied, right? :thinking:

fedellen: Yes, it was added here: [c8d044a](https://github.com/ardriveapp/ardrive-core-js/pull/120/commits/c8d044a0aee8ef13a896bbe8f9877252820e6909)

tests/integration/ardrive.int.test.ts
```diff
@@ -1574,6 +1582,534 @@ describe('ArDrive class - integrated', () => {
...
+			expect(result).to.deep.equal({
+				created: [],
+				tips: [],
+				fees: {}
+			});
+		});
+
+		it('returns the expected revision ArFSResults with a folder that has conflicting names within its tree and uses --upsert conflictResolution on files that return unique different last modified dates', async () => {
+			const conflictingFileInRoot = stubFileToUpload('CONFLICTING_FILE_NAME');
+			stub(conflictingFileInRoot, 'lastModifiedDate').get(() => new UnixTime(1337));
```
matibat, line 1889: I didn't realize I could overwrite values instead of only functions by setting a stubbed getter. Nice approach!

matibat (approved):
It's looking good to me
A very robust test set here! :clap:

PR #124, opened 2022-01-14
PE-880: implement the ArFSDAO methods to send the rename tx

matibat:


src/arfs/arfsdao.ts
```diff
@@ -1397,4 +1398,89 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
+			while (!metaDataUploader.isComplete) {
+				await metaDataUploader.uploadChunk();
+			}
+		}
+
+		return {
+			created: [{ type: 'file', metadataTxId: new TransactionID(metaDataTx.id), entityId: publicFile.entityId }],
+			tips: [],
+			fees: {}
+		};
```
fedellen, line 1440: Usually, we construct this return at the ArDrive layer. I don't really see a problem doing it here instead, but we do need to construct the `fees` information here as well:

```suggestion
		return {
			created: [{ type: 'file', metadataTxId: new TransactionID(metaDataTx.id), entityId: publicFile.entityId }],
			tips: [],
			fees: { [metaDataTx.id]: metaDataTx.reward }
		};
```

fedellen (changes requested):
See comment, we must have the fee information for the metadata tx preserved at this level

src/arfs/arfsdao.ts
fedellen: I just realized we should also be preserving the `dataTxId` information and the `key` information on private renames

matibat: I think this would do the thing: 1c323634ffb220accff6de0706a0f2609221a8a3
Unless the `key` information refers to something else

fedellen: The `key` info in this case is the fileKey. We send it back with private upload-file and also private move-file.

So to remain consistent, in the `created` type on private we should add: `key: urlEncodeHashKey(privateFile.fileKey)`

We also return the `dataTxId` on both public and private files in the same `created` object type: `dataTxId: publicFile.dataTxId`

src/arfs/arfsdao.ts
```diff
@@ -1397,4 +1399,95 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
+		// Upload meta data
+		if (!this.dryRun) {
+			const metaDataUploader = await this.arweave.transactions.getUploader(metaDataTx);
+			while (!metaDataUploader.isComplete) {
+				await metaDataUploader.uploadChunk();
+			}
+		}
+
+		// FIXME: remove this duplicated call
+		const fileKey: FileKey = await deriveFileKey(`${privateFile.fileId}`, driveKey);
```
fedellen, line 1483: Haven't pulled this down, but I believe the fileKey is available on `metadataTxData.fileKey` or `fileMetadata.objectData.fileKey` here.

src/arfs/arfsdao.ts
fedellen: Nice updates! Resolving this

src/arfs/arfsdao.ts
matibat: I've found the file key inside `fileMetadataTxData.fileKey`: 06c6ad2e57c01a081456b4aafdd94f3c5906db80

src/arfs/arfs_entity_result_factory.ts
```diff
@@ -46,6 +46,13 @@ export interface ArFSMoveFileResult extends ArFSMoveEntityResult {
 	dataTxId: TransactionID;
 }
 
+export type ArFSRenameEntityResult = ArFSWriteResult;
+
+export interface ArFSRenameFileResult extends ArFSRenameEntityResult {
```
arielmelendez, line 51: Funny type. It's called a "FileResult" and yet it bears an `EntityType` member... Also the other result data is a `FileID`, even though FileIDs don't change on a rename. Really interested to see how this gets used...

src/arfs/arfsdao.ts
```diff
@@ -1397,4 +1401,96 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 			`The retrieved auth tag does not have the length of ${authTagLength} bytes, but instead: ${data.length}`
 		);
 	}
+
+	async renamePublicFile({
+		fileId,
```
arielmelendez, line 1406: I think this API gets a lot tidier if you just take an ArFSPublicFile. Then you don't need the fileId or owner. Let the ArDrive layer use the DAO to do the fetching.

src/arfs/arfsdao.ts
```diff
@@ -1397,4 +1401,96 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
+		return {
+			type: 'file',
+			entityId: publicFile.fileId,
+			metaDataTxId: new TransactionID(metaDataTx.id),
+			metaDataTxReward: W(metaDataTx.reward)
+		};
+	}
+
+	async renamePrivateFile({
+		fileId,
```
arielmelendez, line 1449: Same here. This gets tidier if you just take an ArFSPrivateFile.

arielmelendez (changes requested):
Left a few design comments here that percolated out to the ArDrive layer in the next PR.

src/arfs/arfs_entity_result_factory.ts
matibat: Cleanup: 1badad2

src/arfs/arfsdao.ts
matibat: Cleanup: 1badad2

src/arfs/arfsdao.ts
matibat: Cleanup: 1badad2

fedellen (approved):

src/arfs/arfsdao.ts
```diff
@@ -1451,6 +1455,88 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
+		if (!this.dryRun) {
+			const metaDataUploader = await this.arweave.transactions.getUploader(metaDataTx);
+			while (!metaDataUploader.isComplete) {
+				await metaDataUploader.uploadChunk();
+			}
+		}
+
+		return {
+			entityId: file.fileId,
+			metaDataTxId: new TransactionID(metaDataTx.id),
```
arielmelendez, line 1492: Nit: You can use the TxID convenience function.

arielmelendez (approved):

src/arfs/arfsdao.ts
matibat: Nice catch! 73a9d1b

PR #125, opened 2022-01-14
PE-881: Implement the ArDrive methods for rename

matibat:


src/ardrive.ts
```diff
@@ -1536,4 +1538,33 @@ export class ArDrive extends ArDriveAnonymous {
...
+		metaDataRewardSettings,
+		driveKey
+	}: RenamePrivateFileParams): Promise<ArFSResult> {
+		if (privateFile.name === newName) {
+			throw new Error(`To rename a file, the new name must be different`);
+		}
+		// this.assertUniqueNameWithinFolder(newName, publicFile.parentFolderId, owner);
+
+		return this.arFsDao.renamePrivateFile(privateFile, newName, metaDataRewardSettings, driveKey);
+	}
```
fedellen, line 1569: This API seems to be a significant change from the rest of the ArDrive layer methods. We're requiring the file entity itself here as well as rewardSettings which we haven't previously exposed.

I would suggest a very simple API, `fileId`, `newName`, and `driveKey`. This would be consistent with the moveFile/moveFolder APIs and would generally be less confusing for a a dev trying to integrate with ArDrive Core.

Also, there seems to be a few steps missing here including:
- Asserting the owner of the drive
- Determining the reward for the transaction
- Asserting the balance of the wallet

src/ardrive.ts
```diff
@@ -1468,6 +1470,24 @@ export class ArDrive extends ArDriveAnonymous {
 		};
 	}
 
+	async estimateAndAssertCostOfFileRename(metaData: ArFSObjectTransactionData): Promise<MetaDataBaseCosts> {
```
fedellen, line 1473: Looks like we can probably re-use the same estimation logic here for both folder creation and move file. If so, we can rename this to something like `estimateAndAssertCostOfMetaDataTx` and excise the other two estimate and assert methods: `estimateAndAssertCostOfFolderUpload` and `estimateAndAssertCostOfMoveFile`.

I would also make this method `private` as I don't think we intend for these estimation methods to be immediately public facing

src/ardrive.ts
```diff
@@ -1536,4 +1556,84 @@ export class ArDrive extends ArDriveAnonymous {
 		const fileToDownload = new ArFSPrivateFileToDownload(privateFile, data, fullPath, decipher);
 		await fileToDownload.write();
 	}
+
+	async renamePublicFile({ fileId, newName, owner }: RenamePublicFileParams): Promise<ArFSResult> {
```
fedellen, line 1560: We could use some integration tests for these new methods. It would provide coverage for both the ArDrive and ArFSDAO layers.

I'd be happy to huddle up tomorrow to get this started

src/ardrive.ts
matibat: Great idea! 5b91164cc7cf4da98f03710799adf64496010fee

src/ardrive.ts
```diff
@@ -1536,4 +1546,102 @@ export class ArDrive extends ArDriveAnonymous {
...
+		}
+	}
+
+	async renamePublicFile({ fileId, newName, owner }: RenamePublicFileParams): Promise<ArFSResult> {
+		await this.assertOwnerAddress(owner);
+		const file = await this.getPublicFile({ fileId, owner });
+		if (file.name === newName) {
+			throw new Error(`To rename a file, the new name must be different`);
+		}
+		await assertValidArFSFileName(newName);
```
arielmelendez, line 1576: no need to await

src/ardrive.ts
```diff
@@ -1536,4 +1546,102 @@ export class ArDrive extends ArDriveAnonymous {
...
+			newName,
+			metadataRewardSettings,
+			owner,
+			driveKey
+		});
+
+		return {
+			created: [
+				{
+					type: result.type,
```
arielmelendez, line 1637: So as pointed out in the other PR, you don't need an entity type in the result type. We know that we are working with a file by way of the APIs we're in.

src/ardrive.ts
```diff
@@ -1536,4 +1546,102 @@ export class ArDrive extends ArDriveAnonymous {
...
+		const allSiblingNames = await this.arFsDao.getPrivateEntityNamesInFolder(folderId, driveKey);
+		const collidesWithExistingSiblingName = allSiblingNames.reduce((accumulator, siblingName) => {
+			return accumulator || siblingName === name;
+		}, false);
+		if (collidesWithExistingSiblingName) {
+			throw new Error(`There already is an entity named that way`);
+		}
+	}
+
+	async renamePublicFile({ fileId, newName, owner }: RenamePublicFileParams): Promise<ArFSResult> {
```
arielmelendez, line 1570: Same comment as in the ArFSFDao PR - make this easier to use correctly and just eliminate the owner from the input params. You've already configured the ArDrive class with the wallet when it was created so the owner shouldn't be allowed to be passed in here. It should be fetched from on chain for the fileId passed in like so:
```
async renamePublicFile({ fileId, newName }: RenamePublicFileParams): Promise<ArFSResult> {
		const owner = await this.arFsDao.getDriveOwnerForFileId(fileId);
```

src/ardrive.ts
```diff
@@ -1536,4 +1546,102 @@ export class ArDrive extends ArDriveAnonymous {
...
+		};
+	}
+
+	async renamePrivateFile({ fileId, newName, owner, driveKey }: RenamePrivateFileParams): Promise<ArFSResult> {
+		await this.assertOwnerAddress(owner);
+		const file = await this.getPrivateFile({ fileId, owner, driveKey });
+		if (file.name === newName) {
+			throw new Error(`To rename a file, the new name must be different`);
+		}
+		await assertValidArFSFileName(newName);
```
arielmelendez, line 1613: no need to await here

arielmelendez (changes requested):
Make APIs easier to use correctly.

src/ardrive.ts
matibat: Going back to here:
 1. The owner of the drive is asserted at the `getPrivateFile` (or public) method
 2. The reward of the transaction is now being returned
 3. There's now a function for asserting the wallet balance before triggering the rename

And the parameters were cleaned up

src/ardrive.ts
matibat: Added at #132

src/ardrive.ts
matibat: I removed the `owner` param; although instead of checking the owner here, it's already enforced at the `getPrivateFile` and `getPublicFile` methods

src/ardrive.ts
matibat: 9fd5eb7

src/ardrive.ts
matibat: 9fd5eb7

fedellen (approved):
Great updates!

src/ardrive.ts
matibat: For the `created` type structure (i.e. `ArFSEntityData[]`) we have we enforce a non-nullable `type` property. So I hardcoded it as `"file"` now.
There's no longer a `result.type`

src/ardrive.ts
arielmelendez: Those methods `getPublic/PrivateFile` FETCH the owner, but do not ENFORCE that the owner is the same as the owner of the configured wallet for the ArDrive class. You will need to enforce the ownership.

arielmelendez (changes requested):
Must enforce ownership

src/ardrive.ts
fedellen: Ah, good catch. Looks like `getPrivateFile` method does enforce the ownership:

```ts
	public async getPrivateFile({ fileId, driveKey, owner }: GetPrivateFileParams): Promise<ArFSPrivateFile> {
		if (!owner) {
			owner = await this.arFsDao.getDriveOwnerForFileId(fileId);
		}

		return this.arFsDao.getPrivateFile(fileId, driveKey, owner);
	}
```

 But not the `getPublicFile` method, as it resides in the anonymous class which does not have a wallet to assert against:
 
 ```ts
 	public async getPublicFile({ fileId, owner }: GetPublicFileParams): Promise<ArFSPublicFile> {
		if (!owner) {
			owner = await this.arFsDao.getDriveOwnerForFileId(fileId);
		}

		return this.arFsDao.getPublicFile(fileId, owner);
	}
 ```

arielmelendez: `getPrivateFile` should not be enforcing ownership. If you have a driveKey, then it doesn't matter if you have the wallet or not. We haven't thought of this as a problem yet, I believe, because we don't allow for creating the ArDrive class without a wallet, and that has trickled its way up to the CLI layer and we have just lived with the consequence on needing a wallet even when we have keys provided (I believe there's a story or defect open to make this issue go away). I'd advise removing the owner assertion from `getPrivateFile` and moving that function down into the anonymous ardrive class as a baby step towards fixing the other ticket. The ownership assertions should happen during write functions only, if we can manage that.

matibat: Let's see here: 4a449566dee00c557250c80be5b88095e7661e9d

arielmelendez (approved):

PR #126, opened 2022-01-14
PE-879: assert unique name before committing the rename

matibat:


src/ardrive.ts
```diff
@@ -1539,32 +1539,59 @@ export class ArDrive extends ArDriveAnonymous {
...
 
-	async renamePublicFile({
-		publicFile,
-		newName,
-		// owner,
-		metaDataRewardSettings
-	}: RenamePublicFileParams): Promise<ArFSResult> {
-		if (publicFile.name === newName) {
+	async assertUniqueNameWithinPublicFolder(name: string, folderId: FolderID, owner: ArweaveAddress): Promise<void> {
+		const allSiblings = await this.listPublicFolder({
```
fedellen, line 1543: Consider `getPublicEntityNamesInFolder` and `getPrivateEntityNamesInFolder` to save on a couple GQL calls.  The list methods are very performant when listing an entire folder, but when we only need the conflicting name information we can do so with a single GQL call

jdaev (approved):
Looks good

src/ardrive.ts
matibat: Here I applied the change: e9e65477c8aa8110223def9f6153f8bc17c7b968
Very helpful comment, thanks!

fedellen (approved):

PR #131, opened 2022-01-18
PE-860: Implement our own getData method with retries

fedellen:


arielmelendez (approved):

src/arfs/arfsdao.ts
```diff
@@ -1270,9 +1271,26 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
+							return exponentialDelay(retryNumber);
+						}
+					});
+					const {
+						data: driveDataBuffer
+					}: {
+						data: Buffer;
+					} = await axiosInstance.get(reqURL, {
+						responseType: 'arraybuffer'
+					});
```
matibat, line 1293: This code is pretty similar to the `getDatForTxID(txId)` method. Does it worth consolidating'em into a method `arweaveGETRequestWithRetries(urlPath)` used by both?

matibat (approved):
Left one single comment, but looks good overall

src/arfs/arfsdao.ts
arielmelendez: Probably yes :) If I have time for it today, I'll give it a whirl. Otherwise would like to get this into QA.

fedellen: Ah, this is in `arfsdao`, was looking for this in the wrong place. Yeah this does seem repetitious here.

PR #136, opened 2022-01-25
PE-896: Implement methods for rename-folder ArDrive

matibat:


src/arfs/arfs_entity_name_validators.ts
```diff
@@ -7,25 +7,28 @@ const CONTAINS_RESERVED_CHARACTER_REGEXP = new RegExp(`[${RESERVED_CHARACTERS.jo
 // From ArFS Standards
 const MAX_VALID_NAME_LENGTH = 255;
 
-export function assertValidArFSFileName(name: string): boolean {
+export const assertValidArFSFileName = assertValidArFSEntityName.bind(this, 'file');
+export const assertValidArFSFolderName = assertValidArFSEntityName.bind(this, 'folder');
+
+export function assertValidArFSEntityName(entityType: 'file' | 'folder', name: string): boolean {
```
arielmelendez, line 13: @karlprieb just a heads up that this will likely be the branch you'll want to branch from for your upcoming ticket.

src/arfs/arfs_entity_name_validators.ts
```diff
@@ -7,25 +7,28 @@ const CONTAINS_RESERVED_CHARACTER_REGEXP = new RegExp(`[${RESERVED_CHARACTERS.jo
 // From ArFS Standards
 const MAX_VALID_NAME_LENGTH = 255;
 
-export function assertValidArFSFileName(name: string): boolean {
+export const assertValidArFSFileName = assertValidArFSEntityName.bind(this, 'file');
```
arielmelendez, line 10: Is binding strictly necessary?

src/ardrive.ts
```diff
@@ -1580,6 +1587,79 @@ export class ArDrive extends ArDriveAnonymous {
...
+		};
+	}
+
+	async renamePrivateFolder({ folderId, newName, driveKey }: RenamePrivateFolderParams): Promise<ArFSResult> {
+		const owner = await this.arFsDao.getDriveOwnerForFolderId(folderId);
+		await this.assertOwnerAddress(owner);
+		const folder = await this.getPrivateFolder({ folderId, driveKey, owner });
+		if (`${folder.parentFolderId}` === ROOT_FOLDER_ID_PLACEHOLDER) {
+			throw new Error(
+				`The root folder cannot be renamed as it shares name with their drive. Try renaming the drive instead`
```
arielmelendez, line 1632:
```suggestion
				`The root folder cannot be renamed as it shares its name with its parent drive. Consider renaming the drive instead.`
```

src/ardrive.ts
```diff
@@ -1580,6 +1587,79 @@ export class ArDrive extends ArDriveAnonymous {
...
+			);
+		}
+		if (folder.name === newName) {
+			throw new Error(`To rename a folder, the new name must be different`);
```
arielmelendez, line 1636:
```suggestion
			throw new Error(`New folder name '${newName}' must be different from the current folder name!`);
```

arielmelendez: Apply this in the public example as well.

src/ardrive.ts
arielmelendez: Apply this in the public example as well.

arielmelendez (changes requested):
Looks good! Just had some minor suggestions around error messages.

src/arfs/arfs_entity_name_validators.ts
karlprieb: 👍 I linked this branch on the ticket description.

src/arfs/arfs_entity_name_validators.ts
matibat: It is not. I'm gonna refactor the function to use the factory pattern instead

src/ardrive.ts
matibat: 922e931a66c06ce8bde2ddd2a4dd90bee3c79cdc

src/ardrive.ts
matibat: 922e931a66c06ce8bde2ddd2a4dd90bee3c79cdc

src/arfs/arfs_entity_name_validators.ts
matibat: Like so: d1db818261c324f0dd3e9092ea5b9852763cf84c

matibat: Added also the missing tests for the `assertValidArFSFolderName` method: 78e710cdc6ffcf25b0b3b7d794daa9e79e93b202

src/arfs/arfs_entity_name_validators.ts
```diff
@@ -7,26 +7,31 @@ const CONTAINS_RESERVED_CHARACTER_REGEXP = new RegExp(`[${RESERVED_CHARACTERS.jo
...
-		throw new Error(`The file name must contain between 1 and ${MAX_VALID_NAME_LENGTH} characters`);
-	}
+export const assertValidArFSFileName = assertValidArFSEntityNameFactory('file');
+export const assertValidArFSFolderName = assertValidArFSEntityNameFactory('folder');
 
-	const matchLeadingSpaces = name.match(LEADING_SPACES_REGEXP);
-	if (matchLeadingSpaces) {
-		throw new Error(`The file name cannot start with spaces`);
-	}
+export function assertValidArFSEntityNameFactory(entityType: 'file' | 'folder'): (name: string) => boolean {
```
arielmelendez, line 13: Is it truly necessary for this to be a factory function? Can it not just take entityType and name as input parameters? My worry is that calling this in a tight loop would thrash the heap while creating and destroying function objects over and over again rather than just using the one loaded into the namespace. Not the end of the world but just seems more efficient to not allocate something new each time.

arielmelendez (approved):

src/arfs/arfs_entity_name_validators.ts
matibat: If you are talking about **multiple instances** of the CLI, then you are right, the factory function will be called three times for each instance. In the other side, if you are talking about the same instance, then we are fine: the funcion is called once per entity type the first time the module is loaded.

PR #138, opened 2022-01-27
PE-938: implement the methods and types for rename-drive at ArFSDAO

matibat:


arielmelendez (approved):

matibat:
> it would be really nice to consolidate all the business logic for all three rename functions using a private member function that uses generics and has injected factory functions

There is PE-936 as a tech debt ticket

karlprieb (approved):

fedellen (approved):
👌

PR #142, opened 2022-01-27
PE-939: add integration tests for rename-drive

matibat:


karlprieb (approved):
We could think about having parameterized tests as the tests are similar for all entities.
It can decrease readability, but we can think about that in the future :smiley:

fedellen (approved):

PR #143, opened 2022-01-30
PE-928: validate drive name when creating new drives

karlprieb:
This PR is heavily based on @matibat work on rename files, folders and drives feature.

- improves ArFS entity name validators tests making them parameterized
- add drive name validation to `createPublicDrive` and `createPrivateDrive` integration tests
- add drive name validation to drive creation methods

PR #145, opened 2022-01-31
PE-930: validate file name when uploading files

karlprieb:
This PR is heavily based on @matibat work on rename files, folders and drives feature and #143 

- improves ArFS entity name validators tests making them parameterized
- add drive file validation to `uploadPublicFIle` and `uploadPrivateFile` integration tests
- add file name validation to file upload method


src/ardrive.ts
```diff
@@ -440,6 +441,8 @@ export class ArDrive extends ArDriveAnonymous {
 			communityTipSettings: CommunityTipSettings
 		) => Promise<ArFSUploadPublicFileResult | ArFSUploadPrivateFileResult>
 	): Promise<ArFSResult> {
+		assertValidArFSFileName(wrappedFile.getBaseFileName());
```
fedellen, line 444: This should be moved below this line and use `destinationBaseName`:

```ts
wrappedFile.newFileName = destinationFileName;

// Assert based on intended destination name
assertValidArFSFileName(wrappedFile.destinationBaseName)

const nameConflictInfo = await getConflictInfoFn(parentFolderId);

```

`getBaseFileName()` will only return the base file name from disk as opposed to the user's intended file name. Sorry for the confusing api there. That one should probably be a private method and the names could be much clearer.

karlprieb: It make sense. Thank you! :+1:

matibat (commented):
Looking good

**Note**
In the future, we might get rid of some of the reserved characters ending only with the Unix ones (e.i. NULL and "/") which are not allowed in the file name. At that moment we might need to stub the `fs` lib to let it return names with reserved characters or maybe dismiss that test.
_This note is not a change request_

src/ardrive.ts
karlprieb: Fixed d88bb917bb2e2e2c15a0ac3a4b5b6ccea710b5aa

fedellen (approved):
👍

src/arfs/arfs_entity_name_validators.test.ts
```diff
@@ -1,4 +1,5 @@
 import { expect } from 'chai';
+
```
arielmelendez, line 2: Whitespace PR spam

arielmelendez (approved):

PR #150, opened 2022-02-03
PE-930: validate names within folder

karlprieb:
This PR is based on #145 

- add `assertNamesWithinFolder` helper function and update unit tests
- add folder and file name validation to `createPublicFolderAndUploadChildren` and `createPrivateFolderAndUploadChildren` and create integration tests for both methods (just testing folder and name validation as they will be deprecated)

### DIFF
--- a/src/arfs/arfs_entity_name_validators.test.ts
+++ b/src/arfs/arfs_entity_name_validators.test.ts
@@ -4,7 +4,7 @@ import { statSync } from 'fs';
 import { ArFSFolderToUpload } from '../arfs/arfs_file_wrapper';
 
 import {
-	assertNamesWithinFolder,
+	assertArFSCompliantNamesWithinFolder,
 	assertValidArFSDriveName,
 	assertValidArFSFileName,
 	assertValidArFSFolderName
@@ -51,7 +51,7 @@ describe('entity name validators', () => {
 	} & Test;
 
 	type FolderWithChildrenTest = {
-		validationMethod: typeof assertNamesWithinFolder;
+		validationMethod: typeof assertArFSCompliantNamesWithinFolder;
 	} & Test;
 
 	function isFolderWithChildren(test: EntityTest | FolderWithChildrenTest): test is FolderWithChildrenTest {
@@ -79,8 +79,8 @@ describe('entity name validators', () => {
 		},
 		{
 			entity: 'folder',
-			methodName: 'assertNamesWithinFolder',
-			validationMethod: assertNamesWithinFolder,
+			methodName: 'assertArFSCompliantNamesWithinFolder',
+			validationMethod: assertArFSCompliantNamesWithinFolder,
 			isFolderWithChildren: true
 		}
 	];
--- a/src/arfs/arfs_entity_name_validators.test.ts
+++ b/src/arfs/arfs_entity_name_validators.test.ts
@@ -1,6 +1,10 @@
 import { expect } from 'chai';
+import { stub } from 'sinon';
+import { statSync } from 'fs';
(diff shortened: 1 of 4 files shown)


src/arfs/arfs_entity_name_validators.test.ts
```diff
@@ -36,64 +40,157 @@ describe('entity name validators', () => {
...
+
+	type EntityTest = {
+		validationMethod: typeof assertValidArFSFileName;
+	} & Test;
+
+	type FolderWithChildrenTest = {
+		validationMethod: typeof assertNamesWithinFolder;
+	} & Test;
+
+	function isFolderWithChildren(test: EntityTest | FolderWithChildrenTest): test is FolderWithChildrenTest {
```
fedellen, line 57: I don't think this type-guard is necessary since you've defined the `isFolderWithChildren` boolean on both types. Below, you could just use the boolean directly: `if (test.isFolderWithChildren)`

Also, I'm not sure if we're saving many lines of code here anymore by trying to fit this into the dynamic tests. Regardless, it the tests seem to be covering the flow just fine

fedellen (approved):
Looking good -- that will cover the names coming from those methods

Thanks for those integration tests!

src/arfs/arfs_entity_name_validators.test.ts
karlprieb: Yep, I tried to check for `test.isFolderWithChildren` and it didn't work as expected, that's why we have this type-guard.

We're not saving more lines, in fact we just added more lines 😄 But we're preventing test setup duplication and ensuring that we're running the same tests for all entity name validation functions.

fedellen: Ah, I see how the compiler would be angry there

arielmelendez: I find this test code hard to follow :( which thus makes assessing it's correctness more difficult to do. I much prefer test code duplication when I am forced to choose between readability/comprehensibility vs. elegance. The elegance doesn't help us if reviewers/code readers can't tell what the tests are doing without deep inspection.

src/arfs/arfs_entity_name_validators.ts
```diff
@@ -38,3 +39,19 @@ export function assertValidArFSEntityNameFactory(entityType: EntityType): (name:
 		return true;
 	};
 }
+
+export function assertNamesWithinFolder(rootFolder: ArFSFolderToUpload, rootFolderDestName?: string): boolean {
+	assertValidArFSFolderName(rootFolderDestName ?? rootFolder.getBaseFileName());
```
arielmelendez, line 44: It worries me that this logic (choosing the name) is duplicated here (it's in about 6 other places in the ArDrive class). It's a type design shortcoming with ArFSFolderToUpload. A built in function in that type that just tells me what the destination name is going to be would be ideal. Not an issue to fix here, but something that I want to broadcast.

src/arfs/arfs_entity_name_validators.ts
```diff
@@ -38,3 +39,19 @@ export function assertValidArFSEntityNameFactory(entityType: EntityType): (name:
 		return true;
 	};
 }
+
+export function assertNamesWithinFolder(rootFolder: ArFSFolderToUpload, rootFolderDestName?: string): boolean {
```
arielmelendez, line 43: Nit: could use a comment explaining what it does, though the name is pretty good :)

arielmelendez: On second thought, the name could be improved: Consider `assertArFSCompliantNamesWithinFolder`

arielmelendez (approved):
Approved. Left some non-mission-critical feedback.

PR #153, opened 2022-02-04
PE-970: Refactor entitiesWithPath

matibat:
~~Implements `ArFSPrivateFileOrFolderWithPathsAndKeys` and its builder in order to have an object representing the entities with paths and keys (only for private ones). The builder is required here to let `ArFSPrivateFileOrFolderWithPathsAndKeys` take an already-constructed `fileKey` (which is derived from an async function at the builder).~~

[See the comment](https://github.com/ardriveapp/ardrive-core-js/pull/153#issuecomment-1034829358)

_Sub-task of PE-771: Add fileKey / driveKey to private list commands_


matibat:
The unit tests for the newly implemented class are at #156 and #157

src/arfs/arfs_entities.ts
```diff
@@ -207,6 +213,19 @@ export class ArFSPrivateFileOrFolderWithPaths extends ArFSFileOrFolderEntity imp
 	}
 }
 
+export class ArFSPrivateFileOrFolderWithPathsAndKeys extends ArFSPrivateFileOrFolderWithPaths implements ArFSWithKeys {
+	readonly driveKey: string;
+	readonly fileKey?: string;
```
arielmelendez, line 218: Really want to avoid this pattern if possible. Handle optionality down in the builders and at higher levels try downcasting to the type that will definitely have the fileKey if needed.

src/arfs/arfs_builders/arfs_file_builders.ts
```diff
@@ -146,6 +147,9 @@ export class ArFSPrivateFileBuilder extends ArFSFileBuilder<ArFSPrivateFile> {
 				case 'Cipher':
 					this.cipher = value;
 					break;
+				case 'File-Key':
+					this.fileKey = Buffer.from(value);
+					break;
```
matibat, line 152: This doesn't belong here. It is removed at #154

src/arfs/arfs_builders/arfs_file_builders.ts
```diff
@@ -25,6 +25,7 @@ interface FileMetaDataTransactionData {
 	lastModifiedDate: number;
 	dataTxId: string;
 	dataContentType: ContentType;
+	fileKey: string;
```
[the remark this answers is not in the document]
matibat: This guy where removed

src/arfs/arfs_entities.ts
```diff
@@ -287,6 +307,10 @@ export class ArFSPrivateFile extends ArFSFileOrFolderEntity {
 	get encryptedDataSize(): ByteCount {
 		return encryptedDataSize(this.size);
 	}
+
+	get fileKey(): FileKey {
```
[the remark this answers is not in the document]
matibat: The getter went away. Now the ArFSPrivateFile will hold its `fileKey` since built

matibat:
### Files with path refactored

Initially, there was `ArFSPublicFileOrFolderWithPaths` as the generic for any File or Folder regardless of its privacy. Now, the requirements specify the need to display the keys associated to each **private** entity.

#### For Public entities
- `ArFSPublicFolderOrFileWithPaths` is the generic for any Public Folder or File as it extends the generic `ArFSFileOrFolderEntity`.

#### For Private entities
- There is a specific `ArFSPrivateFolderWithPaths` which now holds an extra `driveKey` member, and extends the `ArFSFileOrFolderEntity` at the point above.
  - `ArFSPrivateFolderWithPathsKeyless` is a wrapper which hides the `driveKey`.
- And `ArFSPrivateFileWithPaths` extending `ArFSPrivateFolderWithPaths`, which declares another extra member `fileKey`, which is taken at the constructor from the passed ArFSPrivateFile.
  - `ArFSPrivateFileWithPathsKeyless` is a wrapper that hides the `driveKey` and `fileKey`.

#### Changes on entities
- For us to have access to the file keys without having to derivate it each time, the `ArFSPrivateFile` now holds its `fileKey` since build time. Given that some commands print the whole content of this class directly to the user -and that would change the current behavior- a new `ArFSPrivateFileKeyless` was declared for those specific cases.
- No changes were required for the `ArFSPrivateFolder` nor any public class.

src/ardrive_anonymous.ts
```diff
@@ -79,7 +79,7 @@ export class ArDriveAnonymous extends ArDriveType {
 		maxDepth = 0,
 		includeRoot = false,
 		owner
-	}: ListPublicFolderParams): Promise<ArFSPublicFileOrFolderWithPaths[]> {
+	}: ListPublicFolderParams): Promise<ArFSPublicFolderOrFileWithPaths[]> {
```
karlprieb, line 82: I saw that we have two different ways to deal with "FolderOrFileWithPaths", for private one you don't use `ArFSPrivateFileOrFolderWithPaths` anymore, instead you're using `ArFSPrivateFolderWithPaths | ArFSPrivateFileWithPaths`, which I think it's great.
Why not follow the same pattern for `ArFSPublicFolderOrFileWithPaths`?

src/arfs/arfs_entities.ts
```diff
@@ -168,62 +205,72 @@ export class ArFSPublicFileOrFolderWithPaths extends ArFSFileOrFolderEntity impl
...
-		readonly appVersion: string,
-		readonly arFS: string,
-		readonly contentType: ContentType,
-		readonly driveId: DriveID,
-		readonly entityType: EntityType,
-		readonly name: string,
-		readonly txId: TransactionID,
-		readonly unixTime: UnixTime,
-		readonly parentFolderId: FolderID,
+		appName: string,
```
karlprieb, line 259: Why are you removing `readonly`, are we reassigning these values in another place?

matibat: Exactly, that's why: the class inherits these properties from the parent class (e.i. `ArFSFileOrFolderEntity` in this case) which are already read-only. But having in mind this change isn't actually required on the spec, we might want to un-do it :thinking:

src/ardrive_anonymous.ts
matibat: The class were split on the upcoming PR: #156 
At this point, it had no difference because we were using the `entityId` as a generic version of the specific `folderId` and `fileId` of Folder and File respectively. Now separarted.

matibat: Just cherry-picked the commit into here
```
git cherry-pick f916da861f10122d63524cf5bcc9b6cd19e07190
```

src/arfs/arfs_entities.ts
```diff
@@ -139,18 +139,125 @@ export class ArFSFileOrFolderEntity extends ArFSEntity implements ArFSFileFolder
...
+}
+
 export interface ArFSWithPath {
 	readonly path: string;
 	readonly txIdPath: string;
 	readonly entityIdPath: string;
 }
 
-export class ArFSPublicFileOrFolderWithPaths extends ArFSFileOrFolderEntity implements ArFSWithPath {
+export class ArFSPublicFolderWithPaths extends ArFSFileOrFolderEntity {
```
matibat, line 192: Is `WithPath` no longer in use? If so: do we want to make use of it? or delete the interface?

matibat (commented):
**Q** for the reviewer
Should I re-add the `readonly` modifier to the entities constructor argument? That would make the arguments transform into properties (which is already done at the parent classes through inheritance) ovewriting the already set values.

src/arfs/arfs_entities.ts
matibat: We DO want to make use of it.

matibat: Applied here: bbd81edf7140b4cf400a5592808df33a8cdc5d16

src/arfs/arfs_entities.ts
```diff
@@ -139,18 +139,125 @@ export class ArFSFileOrFolderEntity extends ArFSEntity implements ArFSFileFolder
...
+	entity: ArFSPrivateFolder | ArFSPrivateFile,
+	hierarchy: FolderHierarchy,
+	driveKey: DriveKey
+): ArFSPrivateFolderWithPaths | ArFSPrivateFileWithPaths {
+	if (entity.entityType === 'folder') {
+		const entityAsFolder = entity as ArFSPrivateFolder;
+		return new ArFSPrivateFolderWithPaths(entityAsFolder, hierarchy, driveKey);
+	} else if (entity.entityType === 'file') {
+		const entityAsFile = entity as ArFSPrivateFile;
+		return new ArFSPrivateFileWithPaths(entityAsFile, hierarchy, driveKey);
```
fedellen, line 166: Not sure if this makes sense in the flow here, but I think  it makes sense to have `ArFSPrivateFolder` just have the `entityType` field defined as `'folder'` rather than `EntityType` and the same with `ArFSPrivateFile` defining it as `'file'`. 

This would make the `as` casting here unnecessary and the compiler would be able to narrow the types based solely on `entity.entityType === XX`

fedellen: Could be a complicated type battle though with passing the value through the constructors

src/arfs/arfs_builders/arfs_file_builders.ts
fedellen: I don't see this in the diff at #154. Can we just clean it up here?

src/arfs/arfs_entities.ts
matibat: Woah! Thank you so much!
I'm happy with the results: the factories are now much cleaner
84ccf07be40f16c372f49bf05df0c7899587138c

fedellen: Will probably need to sanitize or cast these in the `arfs_builder` layer to account for the type change

matibat: You are right: 006eeb57eeba6a021c16e8b5e703d7785318568c

src/arfs/arfs_builders/arfs_file_builders.ts
```diff
@@ -112,13 +113,13 @@ export class ArFSPublicFileBuilder extends ArFSFileBuilder<ArFSPublicFile> {
...
 	cipherIV?: CipherIV;
+	fileKey?: FileKey;
 
 	constructor(
 		readonly fileId: FileID,
 		readonly arweave: Arweave,
 		private readonly driveKey: DriveKey,
-		readonly owner?: ArweaveAddress,
-		readonly fileKey?: FileKey
+		readonly owner?: ArweaveAddress
```
matibat, line 122: Add `fileKey` back again

matibat, line 122: Add `fileKey` back again
```suggestion
		readonly owner?: ArweaveAddress,
		readonly fileKey?: FileKey
```

fedellen (approved):

src/arfs/arfs_entities.ts
```diff
@@ -139,18 +139,113 @@ export class ArFSFileOrFolderEntity extends ArFSEntity implements ArFSFileFolder
 	}
 }
 
+export function publicEntityWithPathsFactory(
+	entity: ArFSPublicFolder | ArFSPublicFile,
+	hierarchy: FolderHierarchy
+): ArFSPublicFolderWithPaths | ArFSPublicFileWithPaths {
+	if (entity.entityType === 'folder') {
```
arielmelendez, line 146: This is an example of a place where I'd use a switch in case more entity types are added in the future.

src/arfs/arfs_entities.ts
```diff
@@ -139,18 +139,113 @@ export class ArFSFileOrFolderEntity extends ArFSEntity implements ArFSFileFolder
...
+export class ArFSPrivateFolderWithPaths extends ArFSFileOrFolderEntity implements ArFSWithPath {
+	readonly cipher: string;
+	readonly cipherIV: CipherIV;
+	readonly driveKey: string;
 	readonly path: string;
 	readonly txIdPath: string;
 	readonly entityIdPath: string;
 
-	constructor(entity: ArFSPublicFile | ArFSPublicFolder, hierarchy: FolderHierarchy) {
+	constructor(entity: ArFSPrivateFolder, hierarchy: FolderHierarchy, driveKey: DriveKey) {
```
arielmelendez, line 248: I don't like the separation of the driveKey from the ArFSPrivateFolder object here.

src/arfs/arfs_entities.ts
```diff
@@ -166,22 +261,28 @@ export class ArFSPublicFileOrFolderWithPaths extends ArFSFileOrFolderEntity impl
...
 	readonly cipher: string;
 	readonly cipherIV: CipherIV;
+	readonly driveKey: string;
+	readonly fileKey: string;
 	readonly path: string;
 	readonly txIdPath: string;
 	readonly entityIdPath: string;
 
-	constructor(entity: ArFSPrivateFile | ArFSPrivateFolder, hierarchy: FolderHierarchy) {
+	constructor(entity: ArFSPrivateFile, hierarchy: FolderHierarchy, driveKey: DriveKey) {
```
arielmelendez, line 285: Similar to the above, I don't like the separation of the driveKey and fileKey from the ArFSPrivateFile object. As a result you have duplicated key derivation logic below.

src/arfs/arfs_entities.ts
```diff
@@ -197,33 +298,56 @@ export class ArFSPrivateFileOrFolderWithPaths extends ArFSFileOrFolderEntity imp
 			entity.dataTxId,
 			entity.dataContentType,
 			entity.parentFolderId,
-			entity.entityId
+			entity.fileId
 		);
 		this.cipher = entity.cipher;
 		this.cipherIV = entity.cipherIV;
+		this.driveKey = urlEncodeHashKey(driveKey);
```
arielmelendez, line 305: Why is it necessary to urlEncode these on the way in? Perhaps a property getter is necessary that performs that transformation of the raw data on read. As is, this to me is creating the problem of having the same data in 2 places AND is creating confusion for the reader about the expected format of this data.

src/arfs/arfs_entities.ts
arielmelendez: I feel strongly that we need to avoid this if possible.

arielmelendez (changes requested):
Drive key should be stored in the entity type on construction and used elsewhere rather than passing it around separately in many places.

src/arfs/arfs_entities.ts
matibat: I will try instead to add a `toJson` for DriveKey and FileKey. If that creates unwanted conflicts, I'll try adding it to the entities' constructor itself.

The only reason it was represented as a string here is for `JSON.stringify` to do not treat them as a Buffer object

src/arfs/arfs_entities.ts
matibat: This is because I didn't apply this same change to `ArFSPrivateFolder` (as I did for `ArFSPrivateFile` with `fileKey`). I'm gonna apply it as well

src/arfs/arfs_entities.ts
matibat: It seems the typing of Typescript gives us the same safety as `Swift`'s `switch..case`. Here Typescript says: "Alright, `entity` is `ArFSPublicFolder` or `ArFSPublicFile`", then it enters the condition saying: "if this entity has the property `entityType` with value `'folder'`, then it must be an `ArFSPublicFolder`", and when it doesn't enter the condition it says: "hmmm, the entity type of this one is not `'folder'`, then the only left case is `entity` being an `ArFSPublicFile`".

This fabulous behavior came after I applied [the change suggested here](https://github.com/ardriveapp/ardrive-core-js/pull/153#discussion_r804829139) which made the types to be more strict. I.e. previously the builders would create a Folder regardless of what their `entityType` was, now it throws if it mismatches with the **only** allowed value for folders.

src/arfs/arfs_entities.ts
matibat: Applied: [fd475fa](https://github.com/ardriveapp/ardrive-core-js/pull/153/commits/fd475fa775741545ef211d43b9b5280fc65fda44)

src/arfs/arfs_entities.ts
matibat: Applied: [fd475fa](https://github.com/ardriveapp/ardrive-core-js/pull/153/commits/fd475fa775741545ef211d43b9b5280fc65fda44)

src/arfs/arfs_builders/arfs_drive_builders.ts
```diff
@@ -336,7 +338,7 @@ export class SafeArFSDriveBuilder extends ArFSMetadataEntityBuilder<ArFSDriveEnt
 			this.rootFolderId = dataJSON.rootFolderId;
 
 			if (isPrivate) {
-				return new ArFSPrivateDrive(
+				return new ArFSPrivateDriveKeyless(
```
arielmelendez, line 341: I think we could decide whether to return ArFSPrivateDrive vs. ArFSPrivateDriveKeyless based on whether there is a workable driveKey for it (i.e. anything is returned from `this.privateKeyData.driveKeyForDriveId(this.driveId)`) and/or whether the dataJSON matches the placeholder values (i.e. decryption didn't work out).

arielmelendez: My worry about always sending back a keyless one is that you then can't use this API to get back full drives when you do indeed have keys.

src/arfs/arfs_builders/arfs_file_builders.ts
```diff
@@ -80,7 +80,8 @@ export class ArFSPublicFileBuilder extends ArFSFileBuilder<ArFSPublicFile> {
 				this.size === undefined ||
 				!this.lastModifiedDate ||
 				!this.dataTxId ||
-				!this.dataContentType
+				!this.dataContentType ||
+				!(this.entityType === 'file')
```
arielmelendez, line 84: Seems like a good check, but wondering whether it was strictly necessary based on changes made elsewhere?

src/arfs/arfs_entities.ts
```diff
@@ -245,25 +254,57 @@ export class ArFSPublicFile extends ArFSFileOrFolderEntity {
...
-		readonly dataContentType: DataContentType,
+		size: ByteCount,
+		lastModifiedDate: UnixTime,
+		dataTxId: TransactionID,
+		dataContentType: DataContentType,
 		readonly cipher: string,
-		readonly cipherIV: CipherIV
+		readonly cipherIV: CipherIV,
+		readonly fileKey: FileKey,
+		readonly driveKey: DriveKey
```
arielmelendez, line 307: It could make sense to throw in the constructor if the FileKey doesn't match what's expected when the DriveKey and FileID are provided... I'm not sure if it's worth the computational overhead, though, since the builders are already being trusted to provide good values here.

src/arfs/arfs_entities.ts
```diff
@@ -287,6 +328,89 @@ export class ArFSPrivateFile extends ArFSFileOrFolderEntity {
 	get encryptedDataSize(): ByteCount {
 		return encryptedDataSize(this.size);
 	}
+
+	// toJson(): string {
```
arielmelendez, line 332: Should this comment be removed?

src/arfs/arfs_entities.ts
```diff
@@ -287,6 +328,89 @@ export class ArFSPrivateFile extends ArFSFileOrFolderEntity {
...
+			entity.fileKey,
+			entity.driveKey
+		);
+
+		this.path = `${hierarchy.pathToFolderId(entity.parentFolderId)}${entity.name}`;
+		this.txIdPath = `${hierarchy.txPathToFolderId(entity.parentFolderId)}${entity.txId}`;
+		this.entityIdPath = `${hierarchy.entityPathToFolderId(entity.parentFolderId)}${entity.fileId}`;
+	}
+
+	// toJson(): string {}
```
arielmelendez, line 370: Should this comment be removed?

src/arfs/arfs_entities.ts
```diff
@@ -318,25 +442,53 @@ export class ArFSPublicFolder extends ArFSFileOrFolderEntity {
...
+		appVersion: string,
+		arFS: string,
+		contentType: ContentType,
+		driveId: DriveID,
+		readonly entityType: 'folder',
+		name: string,
+		txId: TransactionID,
+		unixTime: UnixTime,
+		parentFolderId: FolderID,
+		readonly folderId: FolderID,
```
arielmelendez, line 488: Don't create memory for folderId.

arielmelendez (changes requested):
Good updates. See my comments.

src/arfs/arfs_builders/arfs_file_builders.ts
matibat: Yes. This came up after ArFSFileEntity had `entityType: 'file'` (instead of `EntityType`). See here: 84ccf07be40f16c372f49bf05df0c7899587138c

src/arfs/arfs_entities.ts
matibat: This should be implemented at `ArFSEntity` (TODO)

src/arfs/arfs_entities.ts
matibat: Same than above

src/arfs/arfs_builders/arfs_drive_builders.ts
matibat: Y should check for the presence of the `driveKey` to conditionally return normal vs keyless

src/arfs/arfs_file_wrapper.ts
```diff
@@ -88,7 +89,7 @@ export class ArFSManifestToUpload extends ArFSEntityToUpload {
 	lastModifiedDateMS: UnixTime;
 
 	constructor(
-		public readonly folderToGenManifest: ArFSPublicFileOrFolderWithPaths[],
+		public readonly folderToGenManifest: (ArFSPublicFolderWithPaths | ArFSPublicFileWithPaths)[],
```
jdaev, line 92: Why would you need to generate a manifest from ArFSPublicFileWithPaths? Isn't it always going to be a folder?

src/arfs/arfs_builders/arfs_drive_builders.ts
arielmelendez: Is that a question? 🧐😬

matibat: Nono, a statement. Thank you :)

src/arfs/arfs_file_wrapper.ts
matibat: I think you are right! @fedellen do you agree?

src/arfs/arfs_entities.ts
```diff
@@ -296,12 +420,12 @@ export class ArFSPublicFolder extends ArFSFileOrFolderEntity {
...
 		readonly contentType: ContentType,
 		readonly driveId: DriveID,
-		readonly entityType: EntityType,
+		readonly entityType: 'folder',
 		readonly name: string,
 		readonly txId: TransactionID,
 		readonly unixTime: UnixTime,
 		readonly parentFolderId: FolderID,
-		readonly entityId: FolderID
+		readonly folderId: FolderID
```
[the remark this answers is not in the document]
matibat: Let's also cache the entity IDs at the constructor so it returns the same instance for the same value. That would in fact also fix the equality operator: by returning the same instance at construction time `new EID(idAsString) === EID(idAsString)`. What do you think?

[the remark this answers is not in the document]
matibat: And for this particular case, we shouldn't care about memory because it is well managed by the EntityID constructor. We will be using pointers to the same data always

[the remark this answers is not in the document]
matibat: Drafting...
```ts
		private static cache: Map<string, EntityID> = new Map();

		// constructor(entityId: string) {
		// const cachedEID = EntityID.cache.get(entityId);
		// if (!cachedEID) {
		// 	EntityID.cache.set(entityId, this);
		// 	return this;
		// }
		// return cachedEID;
```

src/arfs/arfs_entities.ts
matibat: Created a wrapper class for the buffer data of the keys (`EntityKey`) which implements a custom `toJSON` in order for it to be interpreted as the URL-encoded version of the key. [8b4a7d6](https://github.com/ardriveapp/ardrive-core-js/pull/153/commits/8b4a7d61f00503e2f2294d7337d1ba4a9060ebc8)

src/arfs/arfs_entities.ts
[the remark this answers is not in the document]
matibat: As we are passing exactly the same instance to the parent class as given, those are two aliases pointing to the same in-memory data. No extra memory will be reserved in this case.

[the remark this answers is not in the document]
matibat: Also, by using a getter, JSON.stringify wouldn't be able to read it (which may be what we want). What's `readonly` making here is to declare a new attribute.

src/arfs/arfs_file_wrapper.ts
fedellen: This constructor basically takes the output of a `listFolder` and turns it into a valid arweave manifest. To produce the manifest, you will absolutely need the `ArFSPublicFileWithPaths` as well as the target folder's `path`.

I think the variable name could more clearly indicate that.

matibat: So, it takes files and folders with paths: [9fbb5bd](https://github.com/ardriveapp/ardrive-core-js/pull/153/commits/9fbb5bd42a41cca191cf219fc6f47741ef0d5231)

src/arfs/arfs_entities.ts
matibat: Let's have that as a tech_debt ticket if we want the change :)

src/arfs/arfs_entities.ts
matibat: This outdated. Now EntityKey will return a string when stringified

src/ardrive.ts
```diff
@@ -1219,8 +1213,8 @@ export class ArDrive extends ArDriveAnonymous {
 		);
 
 		// Add drive keys to drive and folder entity results
-		createDriveResult.created[0].key = urlEncodeHashKey(newDriveData.driveKey);
-		createDriveResult.created[1].key = urlEncodeHashKey(newDriveData.driveKey);
+		createDriveResult.created[0].key = urlEncodeHashKey(newDriveData.driveKey.keyData);
+		createDriveResult.created[1].key = urlEncodeHashKey(newDriveData.driveKey.keyData);
```
matibat, line 1217:
```suggestion
		createDriveResult.created[0].key = newDriveData.driveKey.toJSON();
		createDriveResult.created[1].key = newDriveData.driveKey.toJSON();
```

src/arfs/arfs_entities.ts
```diff
@@ -289,19 +331,98 @@ export class ArFSPrivateFile extends ArFSFileOrFolderEntity {
...
+			entity.fileKey,
+			entity.driveKey
+		);
+
+		this.path = `${hierarchy.pathToFolderId(entity.parentFolderId)}${entity.name}`;
+		this.txIdPath = `${hierarchy.txPathToFolderId(entity.parentFolderId)}${entity.txId}`;
+		this.entityIdPath = `${hierarchy.entityPathToFolderId(entity.parentFolderId)}${entity.fileId}`;
+	}
+
+	// toJson(): string {}
```
matibat, line 367:
```suggestion
```

tests/stubs.ts
```diff
@@ -96,8 +98,8 @@ export const stubPublicFileMetaDataTx = new ArFSPublicFileMetaDataPrototype(
 	stubEntityIDAltTwo
 );
 
-export const stubPrivateFileMetaDataTx = (async () =>
-	new ArFSPrivateFileMetaDataPrototype(
+export const stubPrivateFileMetaDataTx = (async () => {
+	return new ArFSPrivateFileMetaDataPrototype(
```
matibat, line 102: I guess nothing actually changed here :thinking:

matibat (commented):
A couple of nit comments

src/arfs/arfs_entities.ts
```diff
@@ -139,91 +180,59 @@ export class ArFSFileOrFolderEntity extends ArFSEntity implements ArFSFileFolder
...
+		appVersion: string,
+		arFS: string,
+		contentType: ContentType,
+		driveId: DriveID,
+		readonly entityType: 'file',
+		name: string,
+		txId: TransactionID,
+		unixTime: UnixTime,
+		parentFolderId: FolderID,
 		readonly fileId: FileID,
```
arielmelendez, line 232: In retrospect, fileId should not have its own memory either. It should just be a getter that returns the entityId.

matibat: See my comment [here](https://github.com/ardriveapp/ardrive-core-js/pull/153#discussion_r809155496) about Javascript using pointers to the same instances.

src/ardrive.ts
matibat: 7f6ceec8de9bd872e90ae966c4503101c7de66a7

src/arfs/arfs_entities.ts
matibat: 265837412c0c9e9c941c9a071be9fdebaab4d906

src/types/entity_key.ts
```diff
@@ -0,0 +1,9 @@
+import { urlEncodeHashKey } from '../utils/common';
+
+export class EntityKey {
+	constructor(readonly keyData: Buffer) {}
+
+	toJSON(): string {
```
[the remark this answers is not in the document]
arielmelendez: Yea I think toString makes more sense for the implementation and for toJSON to simply call toString (see EntityID as an example).

src/ardrive.ts
```diff
@@ -1219,8 +1212,8 @@ export class ArDrive extends ArDriveAnonymous {
 		);
 
 		// Add drive keys to drive and folder entity results
-		createDriveResult.created[0].key = urlEncodeHashKey(newDriveData.driveKey);
-		createDriveResult.created[1].key = urlEncodeHashKey(newDriveData.driveKey);
+		createDriveResult.created[0].key = newDriveData.driveKey.toJSON();
```
arielmelendez, line 1215: @karlprieb @fedellen @matibat I think I'd prefer the following; what do you guys this? https://github.com/ardriveapp/ardrive-core-js/pull/165

matibat: Merged down. Thanks!

src/types/entity_key.ts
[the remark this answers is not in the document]
matibat: Applied: [9953305](https://github.com/ardriveapp/ardrive-core-js/pull/153/commits/995330580fcfcc95a65a785de8673c6bbf22ba37)

src/types/entity_key.ts
```diff
@@ -0,0 +1,9 @@
+import { urlEncodeHashKey } from '../utils/common';
+
+export class EntityKey {
+	constructor(readonly keyData: Buffer) {}
```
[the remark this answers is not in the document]
matibat: We do assert keyData is Buffer

matibat (commented):
------

src/types/entity_key.test.ts
```diff
@@ -0,0 +1,66 @@
...
+describe('EntityKey class', () => {
+	let key: EntityKey;
+
+	before(async () => {
+		key = await getStubDriveKey();
+	});
+
+	it('throws if a non-buffer is given', () => {
+		// eslint-disable-next-line prettier/prettier
+		const nonBuffer: Buffer = ('non buffer type' as unknown) as Buffer;
```
arielmelendez, line 14: I feel like if anyone does this, they deserve the hurt that they're going to get at runtime. I don't think we got enough utility out of the new EntityKey class to be honest. I wish we were doing more than checking that someone passed through some kind of crazy type. That itself seems superfluous in typescript world. Instead it'd be cool if we knew of a way to check for the validity of a Buffer. Wrapping the buffer for now is fine though.

src/utils/common.ts
```diff
@@ -377,17 +377,17 @@ export async function createPrivateFileSharingLink(
...
 	let fileSharingUrl = '';
 	try {
-		const driveKey: Buffer = await deriveDriveKey(
+		const driveKey: DriveKey = await deriveDriveKey(
 			user.dataProtectionKey,
 			fileToShare.driveId,
 			user.walletPrivateKey
 		);
-		const fileKey: Buffer = await deriveFileKey(fileToShare.fileId, driveKey);
+		const fileKey: DriveKey = await deriveFileKey(fileToShare.fileId, driveKey);
```
arielmelendez, line 385: Type here should be FileKey, not DriveKey.

arielmelendez (approved):

src/utils/common.ts
matibat: You applied this. Thanks

PR #154, opened 2022-02-04
PE-971: Makes the ArDrive methods to take a withKeys flag

matibat:
Makes `ardrive.listPrivateFolder` to take an optional `withKeys` boolean to decide whether to call `arfsdao.listPrivateFolder` or `arfsdao.listPrivateFolderWithKeys`. This change directly affects `list-drive` also.

**Note** we are temporary having `getPrivateFileKeyless` and `getPrivateFolderKeyless` in order for the `folder-info` and `file-info` to do not change its current behavior. This will be changed in the upcomming task [PE-1027](https://ardrive.atlassian.net/browse/PE-1027) which adds the `--with-keys` parameter to the commands so they behave the same way `list-folder` and `list-drive` does at this PR.

Sub-task of PE-771


src/arfs/arfsdao.ts
```diff
@@ -1344,6 +1345,47 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
 
+	/**
+	 * Lists the children of certain private folder
+	 * @param {FolderID} folderId the folder ID to list children of
+	 * @param {DriveKey} driveKey the drive key used for drive and folder data decryption and file key derivation
+	 * @param {number} maxDepth a non-negative integer value indicating the depth of the folder tree to list where 0 = this folder's contents only
+	 * @param {boolean} includeRoot whether or not folderId's folder data should be included in the listing
+	 * @returns {ArFSPrivateFileOrFolderWithPaths[]} an array representation of the children and parent folder
+	 */
+	async listPrivateFolderWithKeys({
```
fedellen, line 1356: I'm surprised to see this repeating all of the `listPrivateFolder` logic. You could consider using a closure and a shared private implementaion to share the code: 

```ts
private async function someSharedListPrivateFolderFunction<T extends ArFSPrivateFileOrFolderWithPaths>(
	ArFSListPrivateFolderParams,
	entityBuilderFactory: (entity, hierarchy, driveKey) => T
): Promise<T[]>;
```

src/arfs/arfsdao.ts
```diff
@@ -1344,6 +1345,47 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 		return entitiesWithPath;
 	}
 
+	/**
+	 * Lists the children of certain private folder
+	 * @param {FolderID} folderId the folder ID to list children of
+	 * @param {DriveKey} driveKey the drive key used for drive and folder data decryption and file key derivation
+	 * @param {number} maxDepth a non-negative integer value indicating the depth of the folder tree to list where 0 = this folder's contents only
+	 * @param {boolean} includeRoot whether or not folderId's folder data should be included in the listing
+	 * @returns {ArFSPrivateFileOrFolderWithPaths[]} an array representation of the children and parent folder
```
fedellen, line 1354: Update comment or remove.

src/arfs/arfsdao.ts
```diff
@@ -1344,6 +1345,47 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
+		);
+
+		if (includeRoot) {
+			childFolders.unshift(folder);
+		}
+
+		const children = [...childFolders, ...childFiles];
+
+		const entitiesBuilders = children.map(
+			(entity) => new ArFSPrivateFileOrFolderWithPathsAndKeysBuider(entity, hierarchy, driveKey)
```
fedellen, line 1384: spelling nit: `ArFSPrivateFileOrFolderWithPathsAndKeysBuider` -> `ArFSPrivateFileOrFolderWithPathsAndKeysBuilder `

src/arfs/arfsdao.ts
arielmelendez: So echoing this sentiment, my problem with this approach is that it not only duplicates logic, but it also lets values computed and stored in sensible places (i.e. ArFSPrivateFileBuilder, getPrivateFilesWithParentFolderIds, etc.) go to waste. In the current codebase, we have, or are a small step away from deriving the information we need to complete the requirements of this ticket.

I think my preferred approach might be to do something like:
- augment the constructor and membership of ArFSPrivateFile to accept and store the FileKey
- augment the constructor and membership of ArFSPrivateFileMetadataTransactionData to take and store the FileKey
- Remove keys using an Omit<> type at the locations and/or layers where necessary (e.g. ArDrive class or CLI level).

I also noticed that ArFSDao::getPrivateFile doesn't pass along the fileKey when it has the chance (it was added for caching keying purposes). This might be the case in another of other places.

src/arfs/arfsdao.ts
matibat: Removed. Also updated the one at `listPrivateFolder`

src/arfs/arfs_builders/arfs_file_builders.test.ts
```diff
@@ -109,23 +109,28 @@ describe('ArFSPrivateFileBuilder', () => {
 			{ name: 'Parent-Folder-Id', value: 'dde0a0ef-6cd2-45d1-a9b0-97350d9fec21' },
 			{ name: 'File-Id', value: '238e50a9-937b-4160-a3ac-5f8bb0325b70' },
 			{ name: 'Cipher', value: 'AES256-GCM' },
-			{ name: 'Cipher-IV', value: 'nzpsVIcV1vymvT/h' },
+			{ name: 'Cipher-IV', value: 'OqFPdYxI144JrO66' },
```
fedellen, line 112: Was this resulting in a test failure in your environment?

src/ardrive.ts
```diff
@@ -1378,14 +1394,29 @@ export class ArDrive extends ArDriveAnonymous {
 		driveKey,
 		maxDepth = 0,
 		includeRoot = false,
-		owner
-	}: ListPrivateFolderParams): Promise<(ArFSPrivateFolderWithPaths | ArFSPrivateFileWithPaths)[]> {
+		owner,
+		withKeys = false
+	}: ListPrivateFolderParams & { withKeys?: boolean }): Promise<
```
fedellen, line 1399: Could be just added to `ListPrivateFolderParams`

src/arfs/arfs_builders/arfs_file_builders.test.ts
matibat: I was expecting the `fileKey` to be at the file metadataTx data, so this test broke after I wasn't able to read it from there. So I hacked into CLI and re-created the encrypted data to hold it. It is maybe wrong, I'm asking the team :nerd_face:

matibat: Here it is the revert commit: 4917b692b3e2ee8d3c1683c90e9c3dbcdace7a35

src/ardrive.ts
```diff
@@ -1378,14 +1394,29 @@ export class ArDrive extends ArDriveAnonymous {
...
+		}
+
+		const children = this.arFsDao.listPrivateFolder({
+			folderId,
+			driveKey,
+			maxDepth,
+			includeRoot,
+			owner,
+			withPathsFactory
+		});
```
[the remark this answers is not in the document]
matibat: ff380fa1fe6d5a034ae8554f816453691d994345

src/ardrive.ts
matibat: There: 0dfbba8719fb55d97e00c2489052d151da397e1e

src/arfs/arfsdao.ts
```diff
@@ -1310,14 +1310,17 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 	 * @param {DriveKey} driveKey the drive key used for drive and folder data decryption and file key derivation
 	 * @param {number} maxDepth a non-negative integer value indicating the depth of the folder tree to list where 0 = this folder's contents only
 	 * @param {boolean} includeRoot whether or not folderId's folder data should be included in the listing
+	 * @param {ArweaveAddress} owner the arweave address of the wallet which owns the drive
+	 * @param entityWrapper a wrapper used to map the returned entities into
 	 * @returns {ArFSPrivateFileOrFolderWithPaths[]} an array representation of the children and parent folder
```
fedellen, line 1315: Does `ArFSPrivateFileOrFolderWithPaths` exist anymore here?

matibat: It doesn't! :o

src/ardrive.ts
```diff
@@ -1359,6 +1366,15 @@ export class ArDrive extends ArDriveAnonymous {
 		return this.arFsDao.getPrivateFolder(folderId, driveKey, owner);
 	}
 
+	public async getPrivateFileKeyless({
```
arielmelendez, line 1369: Is it really necessary for the DAO to provide this utility when all that needs to be done to get a keyless instance is to wrap an ArFSPrivateFile with the keyless constructor? Just not sure this API has a lot of utility as written. For example, what are its behaviors other than using a known constructor (which exactly matches the return type here)?

matibat: This method was required to avoid the `file-info` CMD changing its behavior after exposing the `fileKey` on the `ArFSPrivateFile` class.
And I've put the method at the `ArDrive` class in order for the CLI layer to do not map the file to its key-less version.

Does it make sense?

src/arfs/arfsdao.ts
matibat: Updated the docblock: 920b96590f9abdf2e73b8452e36d9abe512aa790

matibat (commented):
****

src/ardrive.ts
fedellen: If the current iterations of the `getPrivateFile` and `getPrivateFolder` ArDrive methods have now been changed to include `fileKey` and/or `driveKey` then to remain backwards compatible they need to be adjusted here.

I'd say you could just extend your `withKeys = false` boolean parameter to be applied to the existing private `file-info` and `folder-info` methods and handle that condition rather than defining new API

matibat: There will be a big TODO for this. I created [a task](https://ardrive.atlassian.net/browse/PE-1027) to avoid making this one last longer.

fedellen: Doesn't seem like this one can be a TODO. We're breaking backwards compatibility here by leaking `fileKey` to anyone using `ardrive.getPrivateFile`

matibat: `ArDrive.getPrivateFile` returns the keyless version now, unless you specify `withKeys = true`
ce4b422c959f0a4f61f8cc46fc2a6635e788a4b0

src/ardrive.ts
```diff
@@ -1357,13 +1365,29 @@ export class ArDrive extends ArDriveAnonymous {
...
+		withKeys = false
+	}: GetPrivateFileParams): Promise<ArFSPrivateFile> {
 		if (!owner) {
 			owner = await this.arFsDao.getDriveOwnerForFileId(fileId);
 		}
 		await this.assertOwnerAddress(owner);
 
-		return this.arFsDao.getPrivateFile(fileId, driveKey, owner);
+		const file = await this.arFsDao.getPrivateFile(fileId, driveKey, owner);
+		return withKeys ? new ArFSPrivateFileKeyless(file) : file;
```
arielmelendez, line 1390: This ternary is backwards.

src/arfs/folder_hiararchy.test.ts
```diff
@@ -0,0 +1,22 @@
+describe('FolderHierarchy class', () => {
```
arielmelendez, line 1: I don't like this pattern of creating test files with unimplemented tests. This creates mental overhead for reviewers and in this specific case distracts from the core purpose of the PR.

src/types/ardrive_types.ts
```diff
@@ -168,6 +168,7 @@ export type GetPrivateFolderParams = GetPublicFolderParams & WithDriveKey;
 
 export interface GetPublicFileParams extends GetEntityParams {
 	fileId: FileID;
+	withKeys?: boolean;
```
arielmelendez, line 171: Why would this be in the Public file params? Public files have no keys.

arielmelendez (changes requested):
There's a backwards ternary and a parameter that makes no sense in the public context. Why didn't either of those things cause test failures?

src/arfs/folder_hiararchy.test.ts
matibat: At the previous Retro we emphasized the importance of code changes & tests to share the same PR. :+1:

src/types/ardrive_types.ts
matibat: It was because the private version extends it. Applied only to the private version: [802afb9](https://github.com/ardriveapp/ardrive-core-js/pull/154/commits/802afb95c910d2abd91e50b1136da9f1de661265)

src/arfs/folder_hiararchy.test.ts
karlprieb: Yes, we discussed that, but we should implement the tests not add boilerplate ;)

karlprieb (approved):

PR #156, opened 2022-02-04
PE-981: implements unit tests for FolderHierarchy

matibat:
Sub-task of PE-771


src/arfs/folder_hiararchy.test.ts
```diff
@@ -1,22 +1,169 @@
...
+	let stubHierarchy: FolderHierarchy;
+	let stubHierarchyWithoutRoot: FolderHierarchy;
+
 	before(() => {
-		// TODO: construct a Hierarchy from stubbed entities
+		stubHierarchy = FolderHierarchy.newFromEntities(allFolders);
+		stubHierarchyWithoutRoot = stubHierarchy.subTreeOf(stubEntityIDParent, Number.MAX_SAFE_INTEGER);
+	});
+
+	it('rootNode attribute always return a FolderTreeNode', () => {
```
arielmelendez, line 79: Nit: return -> returns

src/arfs/folder_hiararchy.test.ts
```diff
@@ -1,22 +1,169 @@
+import { expect } from 'chai';
```
arielmelendez, line 1: The file name for this file is misspelled. `hiararchy` -> `hierarchy`

arielmelendez (changes requested):
Fix file name

src/arfs/folder_hiararchy.test.ts
```diff
@@ -1,22 +1,169 @@
...
 	before(() => {
-		// TODO: construct a Hierarchy from stubbed entities
+		stubHierarchy = FolderHierarchy.newFromEntities(allFolders);
+		stubHierarchyWithoutRoot = stubHierarchy.subTreeOf(stubEntityIDParent, Number.MAX_SAFE_INTEGER);
+	});
+
+	it('rootNode attribute always return a FolderTreeNode', () => {
+		const treeRoot = stubHierarchy.rootNode;
+		const subTreeRoot = stubHierarchyWithoutRoot.rootNode;
+		expect(treeRoot).to.be.instanceOf(FolderTreeNode);
```
arielmelendez, line 82: What is the utility of all the type checks in these tests? Why is type the foremost important concept? And why are we not just relying on TypeScript to tell us the types are what we expect?

src/arfs/folder_hiararchy.test.ts
```diff
@@ -1,22 +1,169 @@
...
+		stubEntityIDAltTwo
+	];
+	const stubFolderIDsWithoutRoot = stubFolderIDs.slice(1);
+
+	let stubHierarchy: FolderHierarchy;
+	let stubHierarchyWithoutRoot: FolderHierarchy;
+
 	before(() => {
-		// TODO: construct a Hierarchy from stubbed entities
+		stubHierarchy = FolderHierarchy.newFromEntities(allFolders);
```
arielmelendez, line 75: It's not great from a reader's standpoint that this is called a "stub". It's actually the object under test. What's stubbed are the folders themselves. Consider naming it `folderHierarchy`.

src/arfs/folder_hiararchy.test.ts
```diff
@@ -1,22 +1,169 @@
...
+	 * - The drive name/
+	 * - The drive name/Funny stuff/
+	 * - The drive name/Funny stuff/Funny folder #1/
+	 * - The drive name/Funny stuff/Funny folder #2/
+	 * - The drive name/Funny stuff/Funny folder #3/
+	 * - The drive name/Funny stuff/Funny folder #1/The funniest folder/
+	 */
+
+	// Root folder
+	const stubRootFolder = stubPublicFolder({
```
arielmelendez, line 27: It's a bit annoying that the `stubPublicFolder` function doesn't actually set the folderId that's passed it. It assigns it to the entityId (which is fine given the way that silly ArFSFileOrFolderEntity class is architected) but needs to also follow up after construction and set the folderID. Better yet `ArFSFileOrFolderEntity` should be changed to an abstract class, and the folderId field should be an abstract property that returns the entity ID as null for files and as the entity ID for folders.

src/arfs/folder_hiararchy.test.ts
```diff
@@ -1,22 +1,169 @@
...
+
+	it('rootNode attribute always return a FolderTreeNode', () => {
+		const treeRoot = stubHierarchy.rootNode;
+		const subTreeRoot = stubHierarchyWithoutRoot.rootNode;
+		expect(treeRoot).to.be.instanceOf(FolderTreeNode);
+		expect(subTreeRoot).to.be.instanceOf(FolderTreeNode);
+	});
+
+	it('subTreeOf returns a new FolderHierarchy instance with the correct root node', () => {
+		const subTree = stubHierarchy.subTreeOf(stubEntityIDParent);
```
arielmelendez, line 87: In what sense is this actually evaluating "correctness"?

src/arfs/folder_hiararchy.test.ts
```diff
@@ -1,22 +1,169 @@
...
+		expect(subTreeAlt.rootNode).to.be.instanceOf(FolderTreeNode);
+	});
+
+	it('allFolderIDs returns an array of FolderIDs', () => {
+		const allFolderIDs = stubHierarchy.allFolderIDs();
+		// Ensures the array deeply (but not strictly) has members
+		expect(allFolderIDs).to.have.deep.members(stubFolderIDs);
+		const allFolderIDsWithoutRoot = stubHierarchyWithoutRoot.allFolderIDs();
+		// Ensures the array deeply (but not strictly) has members
+		expect(allFolderIDsWithoutRoot).to.have.deep.members(stubFolderIDsWithoutRoot);
```
arielmelendez, line 101: You should check counts as well since the list in which members are sought might have more items than the output of allFolderIds.

src/arfs/folder_hiararchy.test.ts
```diff
@@ -1,22 +1,169 @@
...
-	it('pathToFolderId returns a path-like string');
-	it('entityPathToFolderId throws if the rootNode of the whole HierarchyTree is not the root of the drive');
-	it('entityPathToFolderId returns a path-like string');
-	it('txPathToFolderId throws if the rootNode of the whole HierarchyTree is not the root of the drive');
-	it('txPathToFolderId returns a path-like string');
+	it('nodeAndChildrenOf returns a flattened array of FolderTreeNode including the given node', () => {
+		const rootNode = stubHierarchy.rootNode;
+		const nodes = stubHierarchy.nodeAndChildrenOf(rootNode, Number.MAX_SAFE_INTEGER);
+		expect(nodes).to.have.lengthOf(6);
+		expect(nodes).to.include(rootNode);
```
arielmelendez, line 108: Why not check that all the nodes are included?

src/arfs/folder_hiararchy.test.ts
```diff
@@ -1,22 +1,169 @@
...
+		nodes.forEach((node) => expect(node).to.be.instanceOf(FolderTreeNode));
+	});
+
+	it('folderIdSubtreeFromFolderId return a flattened array of FolderIDs starting from the given node', () => {
+		const rootNode = stubHierarchy.rootNode;
+		const folderIDs = stubHierarchy.folderIdSubtreeFromFolderId(rootNode.folderId, Number.MAX_SAFE_INTEGER);
```
arielmelendez, line 114: Looks like we're missing tests that actually exercise the depth limitations.

src/arfs/folder_hiararchy.test.ts
matibat: Applied: a340cbe1ac70309e7ce9db4bf52a59c3688c004e

src/arfs/folder_hiararchy.test.ts
matibat: Nice question! Currently, it is only ensuring the returned types are instances of the correct classes. No more than that for now... I might check against the node's property values to assert the correct `folderId` is returned :thinking:

src/arfs/folder_hiararchy.test.ts
matibat: I don't have access to all nodes. Unless I manually instantiate 'em and use the `FolderHierarchy`'s constructor instead of its `newFromEntitires` method.

src/arfs/folder_hiararchy.test.ts
matibat: You are right! Nice catch :sweat_smile:

src/arfs/folder_hiararchy.test.ts
matibat: Applied: 9ff670a0d3b008918c2f7194e79a2017ba133957

src/arfs/folder_hiararchy.test.ts
matibat: Applied: ff2fbe4389009178261ee88f4e518463245f119c

src/arfs/folder_hiararchy.test.ts
matibat: [3050db5](https://github.com/ardriveapp/ardrive-core-js/pull/156/commits/3050db5df584751d25c9c379914f09a04b350b48)

src/arfs/folder_hierarchy.test.ts
```diff
@@ -0,0 +1,207 @@
...
+			expect(() => FolderTreeNode.fromEntity(stubFolder_0)).to.not.throw();
+		});
+
+		it('Can be constructed from a folderId', () => {
+			expect(() => new FolderTreeNode(stubEntityID)).to.not.throw();
+			expect(() => new FolderTreeNode(stubEntityID, folderHierarchy.rootNode)).to.not.throw();
+		});
+
+		describe('the maxDepth parameter', () => {
+			it('the sub-tree of maxDepht zero has only one node on it', () => {
```
arielmelendez, line 179: maxDepht -> maxDepth

arielmelendez, line 179: This test is failing for me locally.

matibat: Yes! `FolderHierarchy` has two bugs.
 - One es fixed at this PR (bb028db3295aea2affefde6a2fbc6537da82832f)
 - and the other is a known defect [PE-1012](https://ardrive.atlassian.net/browse/PE-1012)

matibat: Ty: 78dc7b808a28892612c64c99244757efd3ae2f4c

matibat: I'll skip that test and leave there a comment indicating the reason (PE-1012)

matibat: 1b21f2bc6c116d1fd9e289fbea3eb8f6d78ece0f

karlprieb (approved):

src/arfs/folder_hierarchy.test.ts
```diff
@@ -0,0 +1,205 @@
...
+		describe('the maxDepth parameter', () => {
+			it.skip('the sub-tree of maxDepth zero has only one node on it', () => {
+				// Skipped until PE-1012 gets done
+				const subTree = folderHierarchy.subTreeOf(stubRootFolder.folderId, 0);
+				const rootNode = subTree.rootNode;
+				const allNodesOfSubTree = subTree.nodeAndChildrenOf(rootNode, Number.MAX_SAFE_INTEGER);
+				expect(allNodesOfSubTree.length, `Wrong sub-tree length!`).to.equal(1);
+			});
+
+			it('maxDepth of zero', () => {
```
arielmelendez, line 187: These test names are not to the standards we expect. As written, this test name would be:

`FolderHierarchy class FolderTreeNode class the maxDepth parameter maxDepth of zero`

This test name does not:
- explain very clearly which class is under test
- explain very clearly which function of that class is under test
- explain very clearly what the inputs and/or conditions of the test are
- explain very clearly what the behaviors and outputs of the object/function under test are as a result.

When a test lacks one or more of the above, it's very difficult for a reader to assess whether the appropriate test coverage is being applied.

When I look at the body of this test, what I see being tested is:
- 'FolderHierarchy::nodeAndChildrenOf' being tested with some FolderTreeNode and a max depth of 0 (note: it's possible that this function could be used with a FolderTreeNode that's not even in the hierarchy. That points to questionable class design. Maybe this should be a static FolderTreeNode function instead since no private state of either FolderHierarchy or FolderTreeNode is used?) and making sure that the returned list is of the right length, but NOT checking whether it's made of the right nodes (this is a shortcoming).
- 'FolderHierarchy:: folderIdSubtreeFromFolderId' is being tested with a known good folderId (by the way, that function will probably crash if a bad folderId is passed... maybe not the best type and/or function design...) and a max depth of 0. Again, it tests that it the returned list is of the right length, but does NOT check whether it's made of the right nodes (this is, again, a shortcoming).

So in summary:
- The test name is not indicative of what's actually being tested
- Two functions (i.e. UNITS) are under test instead of one
- The inputs aren't clearly/fully articulated in the test name
- The expected outputs are not articulated at all in the test name
- The outputs are only partially verified.

arielmelendez (approved):
These tests are not everything that we'd want, but are better than what we had (which was nothing). Given the time invested thus far I think it's time for us to move on.

PR #157, opened 2022-02-04
PE-981: implements the pending tests for the entities with key builder

matibat:
Sub task of PE-771


fedellen (approved):

karlprieb (approved):

fedellen (approved):

src/arfs/arfs_builders/arfs_with_keys.test.ts
```diff
@@ -0,0 +1,115 @@
...
+	privateEntityWithPathsKeylessFactory
+} from '../../exports';
+import { DriveKey, FileKey } from '../../types';
+import { urlEncodeHashKey } from '../../utils/common';
+import { RootFolderID } from './arfs_folder_builders';
+
+describe('Entities with paths classes', () => {
+	/*
+	 * The stubbed folders describes the following tree
+	 * - The drive name/
```
arielmelendez, line 26: These names are fun, but for the sake of readability and comprehensibility, something like the following might have been more appropriate:
```
- Root folder/Parent Folder/Child Folder
```

src/arfs/arfs_builders/arfs_with_keys.test.ts
```diff
@@ -0,0 +1,115 @@
...
+	 * - The drive name/Funny stuff/Funny folder #1/jokes.pdf
+	 */
+
+	// Root folder
+	const stubRootFolder = stubPrivateFolder({
+		folderId: stubEntityIDRoot,
+		parentFolderId: new RootFolderID(),
+		folderName: 'The drive name'
+	});
+	// Depth 0
```
arielmelendez, line 40: These depth labels might be confusing since the root folder is actually depth 0.

arielmelendez (approved):
These tests are fine but where are the integration tests for the ardrive class?

src/arfs/arfs_builders/arfs_with_keys.test.ts
matibat: Cleaned it up

src/arfs/arfs_builders/arfs_with_keys.test.ts
matibat: Applied

karlprieb:
> These tests are fine but where are the integration tests for the ardrive class?

I have the same question from @arielmelendez

tests/integration/ardrive.int.test.ts
```diff
@@ -194,6 +205,293 @@ describe('ArDrive class - integrated', () => {
...
+				);
+				stub(arfsDao, 'getDriveOwnerForFolderId').resolves(walletOwner);
+				stub(arfsDao, 'getPublicFolder').resolves(stubPublicRootFolder);
+			});
+
+			describe('maxDepth parameter', () => {
+				it('throws if provided a negative value', () => {
+					expectAsyncErrorThrow({
+						promiseToError: arDrive.listPublicFolder({ folderId: stubEntityIDRoot, maxDepth: -1 }),
+						errorMessage: 'Me gustan las empanadas'
```
matibat, line 234: Clearly, you love empanadas but why isn't this mismatching the error message?

matibat (commented):
--------

karlprieb (approved):

tests/integration/ardrive.int.test.ts
```diff
@@ -194,6 +205,293 @@ describe('ArDrive class - integrated', () => {
...
+					});
+				});
+
+				it('lists with depth zero by default', async () => {
+					const entities = await arDrive.listPublicFolder({ folderId: stubEntityIDRoot });
+					expect(entities.length).to.equal(2);
+					expect(entities).to.deep.equal(
+						[stubPublicParentFolder, stubPublicFileInRoot].map((entity) =>
+							publicEntityWithPathsFactory(entity, stubPublicHierarchy)
```
arielmelendez, line 250: You really should use as little code from the inner implementation of the function under test as possible as part of your test code, otherwise you're vulnerable to false positives. It is like "the fox guarding the hen house". If you know exactly what the entities are, just put them in a literal list like:
```
expect(entities).to.deep.equal([
    new ArFSPublicFolderWithPaths(stubPublicParentFolder, stubPublicHierarchy),
    new ArFSPublicFolderWithPaths(stubPublicFileInRoot, stubPublicHierarchy)
])
```

arielmelendez (approved):
Thanks for adding the expanded tests. It's risky to use code under test in your validations, but we can move forward as is for now.

PR #158, opened 2022-02-04
PE-983: Update entity name validation rules

karlprieb:
It removes some entity name rules. As we have less rules I thought it would be good to add those 3 invalid name cases to all our integration tests.


src/arfs/arfs_entity_name_validators.test.ts
```diff
@@ -91,17 +90,17 @@ describe('entity name validators', () => {
 		const { entity, methodName } = test;
 
 		describe(`${methodName} method`, () => {
-			it('returns true when fed with an ArFS compliant name', () => {
+			it('do not throws when fed with an ArFS compliant name', () => {
```
fedellen, line 93: nit:
```suggestion
			it('does not throw an error when fed with an ArFS compliant name', () => {
```

fedellen (approved):

PR #161, opened 2022-02-07
PE-480: Custom gateway flag

fedellen:
This PR ensures that the gateway supplied from the `arweave` instance supplied by the `ardriveFactory` is properly seeded down and used in each layer. All hard coded gateway URLs have been removed in favor of deriving it from `this.arweave` instance

To test this at the Core level, ArLocal was added to the repo. I've added an ArLocal integration test file with a basic set of happy paths for ArDrive write methods: `createDrive`, `createFolder`, `uploadFile`, and `createFolderAndUploadChildren`  (Public/Private). The tests also read those entities back from ArLocal testing these methods: `getFile`, `getFolder`, `getDrive`, and `listFolder`  (Public/Private).


src/utils/common.ts
```diff
@@ -541,11 +542,20 @@ export function readJWKFile(path: string): Wallet {
 	return wallet;
 }
 
-export async function fetchMempool(): Promise<string[]> {
-	const response = await axios.get('https://arweave.net/tx/pending');
+export async function fetchMempool(gateway = new URL('https://arweave.net/')): Promise<string[]> {
```
arielmelendez, line 545: Can we factor out `https://arweave.net/` as a constant somewhere?

arielmelendez: Or use an existing one.

src/utils/common.ts
```diff
@@ -541,11 +542,20 @@ export function readJWKFile(path: string): Wallet {
...
+	const response = await axios.get(`${gateway.href}tx/pending`);
 	return response.data;
 }
 
+/** Derives gateway URL from provided Arweave instance */
+export function gatewayUrlForArweave(arweave: Arweave): URL {
+	const protocol = arweave.api.config.protocol ?? 'https';
```
arielmelendez, line 552: Constants over literals

src/utils/constants.ts
```diff
@@ -10,10 +10,6 @@ export { DEFAULT_APP_VERSION };
 
 export const prodAppUrl = 'https://app.ardrive.io';
 export const stagingAppUrl = 'https://staging.ardrive.io';
-export const gatewayURL = 'https://arweave.net/';
```
arielmelendez, line 13: Add this back as a default in service of the small handful of places we're using them.

src/arfs/arfsdao.ts
```diff
@@ -1040,7 +1039,7 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 				owner
 			});
 
-			const response = await this.arweave.api.post(graphQLURL, gqlQuery);
+			const response = await this.arweave.api.post('graphql', gqlQuery);
```
arielmelendez, line 1042: Prefer a constant over a literal

src/arfs/arfsdao_anonymous.ts
```diff
@@ -100,7 +100,7 @@ export class ArFSDAOAnonymous extends ArFSDAOType {
 					],
 					sort: ASCENDING_ORDER
 				});
-				const response = await this.arweave.api.post(graphQLURL, gqlQuery);
+				const response = await this.arweave.api.post('graphql', gqlQuery);
```
arielmelendez, line 103: Prefer constant over literal

src/pricing/gateway_oracle.ts
```diff
@@ -4,8 +4,10 @@ import { BigNumber } from 'bignumber.js';
 import axios, { AxiosResponse } from 'axios';
 
 export class GatewayOracle implements ArweaveOracle {
+	constructor(private readonly gateway = new URL('https://arweave.net/')) {}
```
arielmelendez, line 7: Use constant over literal

src/arfs/arfs_builders/arfs_builders.ts
```diff
@@ -74,7 +74,7 @@ export abstract class ArFSMetadataEntityBuilder<T extends ArFSEntity> {
 		if (!node) {
 			const gqlQuery = buildQuery({ tags: this.getGqlQueryParameters(), owner });
 
-			const response = await this.arweave.api.post(graphQLURL, gqlQuery);
+			const response = await this.arweave.api.post('graphql', gqlQuery);
```
arielmelendez, line 77: use constant over literal (just search everywhere for this string and do the same)

package.json
```diff
@@ -69,8 +69,9 @@
 		"power-assert": "mocha --parallel false --r espower-typescript/guess ./**/*test.ts",
 		"typecheck": "tsc --noemit",
 		"build": "yarn clean && tsc --project ./tsconfig.prod.json",
-		"ci": "yarn test && yarn build",
-		"dev": "yarn clean && tsc --project ./tsconfig.prod.json -w"
+		"ci": "yarn arlocal-docker-test && yarn build",
+		"dev": "yarn clean && tsc --project ./tsconfig.prod.json -w",
+		"arlocal-docker-test": "docker run --init --rm -d -p 1984:1984 textury/arlocal:latest && yarn nyc mocha && docker stop $(docker ps -q)"
```
karlprieb, line 72: maybe we can set a name to docker container, like `docker run --init --rm -d -p 1984:1984 --name arlocal textury/arlocal:latest` so que can use `docker stop arlocal` instead of stopping every container.

tests/integration/arlocal.int.test.ts
```diff
@@ -38,22 +38,24 @@ import {
 	ArFSPublicFolder
 } from '../../src/arfs/arfs_entities';
 
-describe('ArLocal Integration Tests', () => {
+describe('ArLocal Integration Tests', function () {
+	this.timeout(5000);
```
karlprieb: Instead of increasing the test timeout we can use `before` hook with `fetch` to localhost:1984 (maybe increase fetch timeout?) so we ensure that we only start tests when arlocal is up.

karlprieb: I saw that you don't have this timeout anymore, it was not needed?

fedellen: I ran into a timeout on `uploadPublicFile` here on a couple runs, doesn't seem to timeout on CI. I should actually do a little profiling there and see where its hanging. It doesn't happen on `uploadPrivateFile`

package.json
fedellen: Yeah that makes more sense. We don't want to arbitrarily stop the BATS docker container when we're running core tests for example

I'm going to try to push this into a bash script so we can run ArLocal tests with grep and so the stop script is more reliable

fedellen: Included this with the bash script: [e93cf96](https://github.com/ardriveapp/ardrive-core-js/pull/161/commits/e93cf9634f2994dfdd17676575916cd627dc5f44)

src/arfs/arfs_builders/arfs_builders.ts
fedellen: [324e705](https://github.com/ardriveapp/ardrive-core-js/pull/161/commits/324e7056276a01ffb686fbe2fba958b130dcbffc)

src/pricing/gateway_oracle.ts
fedellen: [324e705](https://github.com/ardriveapp/ardrive-core-js/pull/161/commits/324e7056276a01ffb686fbe2fba958b130dcbffc)

src/arfs/arfsdao.ts
fedellen: [324e705](https://github.com/ardriveapp/ardrive-core-js/pull/161/commits/324e7056276a01ffb686fbe2fba958b130dcbffc)

src/utils/common.ts
fedellen: [324e705](https://github.com/ardriveapp/ardrive-core-js/pull/161/commits/324e7056276a01ffb686fbe2fba958b130dcbffc)

src/utils/common.ts
fedellen: [324e705](https://github.com/ardriveapp/ardrive-core-js/pull/161/commits/324e7056276a01ffb686fbe2fba958b130dcbffc)

karlprieb (approved):
👍

arielmelendez (approved):

tests/arlocal-docker-test.sh
```diff
@@ -0,0 +1,11 @@
...
+
+docker run --init --rm -d -p 1984:1984 --name arlocal textury/arlocal:latest
+
+# Add grep support for arlocal mocha tests
+if [ -z "$1" ];
+then yarn nyc mocha; 
+else yarn nyc mocha -g "$1"; 
+fi
+
+docker stop arlocal
```
arielmelendez, line 11: I think you need to use set -e and capture the return code from the yarn invocation and forward it out of this script or else you might leave containers running. Here's how to do that:

https://stackoverflow.com/a/45729843

arielmelendez (changes requested):
Don's stop the script if yarn errors.

package.json
```diff
@@ -67,8 +67,9 @@
 		"power-assert": "mocha --parallel false --r espower-typescript/guess ./**/*test.ts",
 		"typecheck": "tsc --noemit",
 		"build": "yarn clean && tsc --project ./tsconfig.prod.json",
-		"ci": "yarn test && yarn build",
-		"dev": "yarn clean && tsc --project ./tsconfig.prod.json -w"
+		"ci": "yarn arlocal-docker-test && yarn build",
+		"dev": "yarn clean && tsc --project ./tsconfig.prod.json -w",
+		"arlocal-docker-test": "\"tests/arlocal-docker-test.sh\""
```
matibat, line 72: I haven't tried it yet, but `tests/arlocal-docker-test.sh` might be missing execution permissions just after cloning. I that case, this would do the trick:
```sh
bash "tests/arlocal-docker-test.sh"
```
_You can ignore if this is not an issue_

tests/integration/arlocal.int.test.ts
```diff
@@ -0,0 +1,747 @@
...
+	ArFSEntity,
+	ArFSPrivateDrive,
+	ArFSPrivateFile,
+	ArFSPrivateFileOrFolderWithPaths,
+	ArFSPrivateFolder,
+	ArFSPublicDrive,
+	ArFSPublicFile,
+	ArFSPublicFileOrFolderWithPaths,
+	ArFSPublicFolder
+} from '../../src/arfs/arfs_entities';
```
matibat, line 39: I might end up updating the WithPaths entities at this file when I merge down PE-771

matibat (dismissed):
I'm good with the changes

tests/arlocal-docker-test.sh
fedellen: Is it something like this? 

3ee9df3

package.json
fedellen: Nice catch! Updated this just in case: 28eb0b8

fedellen:
Changed the base branch here to include the PE-460 / PE-795 integration branch.

tests/arlocal-docker-test.sh
```diff
@@ -0,0 +1,15 @@
...
+EXIT_CODE=0
+
+# Add grep support for arlocal mocha tests
+if [ -z "$1" ];
+then yarn nyc mocha || EXIT_CODE=$?; 
+else yarn nyc mocha -g "$1" || EXIT_CODE=$?; 
+fi
+
+docker stop arlocal
+echo $EXIT_CODE
```
karlprieb, line 15: Why would we want to `echo` the exit code?
Maybe we want `exit $EXIT_CODE`?

fedellen: Sounds good to me: 4bb2753

jdaev (commented):
Looks good to me 👍🏽

matibat (approved):
LGTM

karlprieb (approved):
👍

arielmelendez (approved):

PR #162, opened 2022-02-09
PE-795: file name byte length

karlprieb:
Change the validation for max file name character length to max 255 bytes
Update all integration tests using `assertEntityNameExpectations` helper
Alphabetical sorting on imports

### DIFF
--- a/src/arfs/arfs_entity_name_validators.ts
+++ b/src/arfs/arfs_entity_name_validators.ts
@@ -2,17 +2,23 @@ import { EntityType } from '../types';
 import { ArFSFolderToUpload } from '../arfs/arfs_file_wrapper';
 
 // From ArFS Standards
-const MAX_VALID_NAME_LENGTH = 255;
+const MAX_VALID_NAME_BYTE_LENGTH = 255;
 
 export const assertValidArFSFileName = assertValidArFSEntityNameFactory('file');
 export const assertValidArFSFolderName = assertValidArFSEntityNameFactory('folder');
 export const assertValidArFSDriveName = assertValidArFSEntityNameFactory('drive');
 
 export function assertValidArFSEntityNameFactory(entityType: EntityType): (name: string) => void | Error {
-	return function (name: string) {
-		// Check for max length and empty names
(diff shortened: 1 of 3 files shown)


fedellen (approved):
Looks good to me!

src/arfs/arfs_entity_name_validators.ts
```diff
@@ -2,17 +2,23 @@ import { EntityType } from '../types';
...
-			throw new Error(`The ${entityType} name must contain between 1 and ${MAX_VALID_NAME_LENGTH} characters`);
+	return function (name: string): void {
+		// Check for empy names
+		if (name.length === 0) {
+			throw new Error(`The ${entityType} name cannot be empty`);
+		}
+
+		// Check for max byte length
+		const nameByteLength = new TextEncoder().encode(name).length;
+		if (nameByteLength > MAX_VALID_NAME_BYTE_LENGTH) {
```
fedellen, line 20: Wait, isn't this supposed to be `nameByteLength >= MAX_VALID_NAME_BYTE_LENGTH`

fedellen: Whereas a name with with 255 bytes would error

fedellen (changes requested):
Behavior doesn't seem to match the error message

src/arfs/arfs_entity_name_validators.ts
fedellen: We should add some boundary test cases here, one for a 254 byte name and one for a 255 byte name. The current tests seems to only test for a 256 character name (presumably 256 bytes)

karlprieb: In fact most *nix FS supports 255 bytes max.
On unit test we have several valid file names, the second one is one with 255 bytes.

karlprieb: Upated error mesage [e1e8944](https://github.com/ardriveapp/ardrive-core-js/pull/162/commits/e1e8944549cea4725c4ad2ee908f1745587954be)

fedellen (approved):

PR #164, opened 2022-02-17
Pe 970 cleanup

arielmelendez:


src/types/types.ts
```diff
@@ -13,6 +13,7 @@ export type FolderID = EntityID;
 export type FileID = EntityID;
 export type DriveID = EntityID;
 export type AnyEntityID = DriveID | FolderID | FileID;
+export type EntityIDTypeForEntityType<T extends 'file' | 'folder'> = T extends 'file' | 'folder' ? FileID : FolderID;
```
arielmelendez, line 16: Could behave unexpectedly for an actual union type passed through...

arielmelendez:
Double check all ArFSFileOrFolderEntity type references that don't include a file or folder specification to make sure they're being used safely.

src/types/types.ts
matibat: I've found at an example given [here](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html) that the type
```ts
TypeToID = T extends 'file' ? FileID : FolderID
```
Will behave the following way:
```ts
function createLabel<T extends 'file | 'folder>(type: T): TypeToID<T> {
  throw "unimplemented";
}

const folderId = createLabel('folder'); // Type FolderID
const fileId = createLabel('file'); // Type FileID

let folderOrFile: 'file' | 'folder' = 'file'
const folderOrFileId = createLabel(folderOrFile); // Type FileID | FolderID
```

matibat: Applied the change: 0cbe0d7dd199f57e6d582f2fab80095c3585c992

src/arfs/arfs_builders/arfs_builders.ts
```diff
@@ -152,7 +152,10 @@ export abstract class ArFSMetadataEntityBuilder<T extends ArFSEntity> {
 	}
 }
 
-export abstract class ArFSFileOrFolderBuilder<T extends ArFSFileOrFolderEntity> extends ArFSMetadataEntityBuilder<T> {
+export abstract class ArFSFileOrFolderBuilder<
+	U extends 'file' | 'folder',
+	T extends ArFSFileOrFolderEntity<U>
+> extends ArFSMetadataEntityBuilder<T> {
```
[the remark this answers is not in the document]
arielmelendez: I don't think this gives us the co-variance we need for ArFSFileOrFolderEntity. I believe we tried this and it didn't work in handful of use cases.

PR #165, opened 2022-02-18
Pe 970 arfsresult update

arielmelendez:


src/types/ardrive_types.ts
```diff
@@ -25,7 +26,7 @@ export interface ArFSEntityData {
 	metadataTxId?: TransactionID;
 	dataTxId?: TransactionID;
 	entityId?: AnyEntityID;
-	key?: string;
+	key?: EntityKey;
```
matibat, line 29: I like this change. Applied at: af77bce9346b063c26e871825e63070ac1e2e6f7

matibat (approved):
I didn't see this PR but applied half of the changes to the base PR. The only difference I see is using 
```ts
key.toString()
````
vs using 
```ts
`${key}`
```
So merging the changes would be functionally similar :+1:

PR #166, opened 2022-02-22
PE-771: Add fileKey / driveKey to private list commands

matibat:
Integration branch of all sub-tasks


src/ardrive.ts
```diff
@@ -547,7 +550,7 @@ export class ArDrive extends ArDriveAnonymous {
 				type: 'folder',
 				entityId: folderId,
 				metadataTxId: folderTxId,
-				key: driveKey ? urlEncodeHashKey(driveKey) : undefined
+				key: driveKey ? driveKey : undefined
```
karlprieb, line 553: I know that can be late to change, but we could use `key: driveKey ?? undefined`.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

src/types/entity_key.test.ts
```diff
@@ -0,0 +1,66 @@
...
+				// eslint-disable-next-line prettier/prettier
+				189,
+				19,
+				169,
+				3,
+				242,
+				227,
+				175,
+				155,
+				172
```
fedellen, line 54: Seems like the formatter ignored these eslint-disables

src/utils/common.ts
```diff
@@ -28,13 +28,13 @@ export async function sleep(ms: number): Promise<number> {
...
 // Asyncronous ForEach function
-export async function asyncForEach(array: any[], callback: any): Promise<string> {
+/*export async function asyncForEach(array: any[], callback: any): Promise<string> {
 	for (let index = 0; index < array.length; index += 1) {
 		// eslint-disable-next-line no-await-in-loop
 		await callback(array[index], index, array);
 	}
 	return 'Done';
-}
+}*/
```
fedellen, line 37: Deleting this would be fine. We can cleanup the `common.ts` file later on though

tests/stubs.ts
```diff
@@ -402,27 +410,31 @@ const stubSpecialCharFileInChild = stubPublicFile({
...
-export const stubSpecialCharEntitiesWithPaths = stubSpecialCharEntities.map(
-	(entity) => new ArFSPublicFileOrFolderWithPaths(entity, stubSpecialCharHierarchy)
+export const stubSpecialCharEntitiesWithPaths = stubSpecialCharEntities.map((entity) =>
+	publicEntityWithPathsFactory(entity, stubSpecialCharHierarchy)
 );
 
 export const stubEntitiesWithOneFile = [stubPublicRootFolder, stubPublicFileInRoot];
 
-export const stubHierarchyWithOneFile = FolderHierarchy.newFromEntities(stubEntitiesWithOneFile);
+export const stubHierarchyWithOneFile = FolderHierarchy.newFromEntities([stubPublicRootFolder]);
```
fedellen, line 434: This change doesn't make sense to me. `stubHierarchyWithOneFile` only has a public folder with no file?

fedellen: It would probably be failing some test cases if something was wrong though

src/ardrive.ts
matibat: Aggree

src/types/entity_key.test.ts
matibat: Git blames @arielmelendez ahahaha
I'm gonna try to un-do it

src/utils/common.ts
matibat: I can do it quickly

src/ardrive.ts
```diff
@@ -912,30 +915,75 @@ export class ArDrive extends ArDriveAnonymous {
...
 			owner = await this.getOwnerForDriveId(driveId);
 		}
 		await this.assertOwnerAddress(owner);
 
-		return this.arFsDao.getPrivateDrive(driveId, driveKey, owner);
+		const drive = await this.arFsDao.getPrivateDrive(driveId, driveKey, owner);
+		return withKeys
+			? drive
+			: new ArFSPrivateDriveKeyless(
+					drive.appName,
```
karlprieb, line 933: We need to create a tech debt for `ArFSPrivateDriveKeyless` constructor to expect the entity instead of all those properties :)

src/ardrive.ts
matibat: Actually being that driveKey has type `EntityKey | undefined`, `driveKey ?? undefined` is logically the same than `driveKey`

tests/stubs.ts
fedellen: Ah, maybe the method name should clearly indicate that. Like `newFromFolderEntities` or similar

karlprieb (approved):

src/types/entity_key.test.ts
matibat: 068230038ff60a6d52341f50686c13146c583a0f

karlprieb (approved):

fedellen (approved):

tests/stubs.ts
matibat: At least now the types are more strict and won't let you put a file on there

PR #167, opened 2022-02-23
PE-1038: Release ArDrive Core v1.8.0

fedellen:
This PR releases ArDrive Core v1.8.0 with story tickets: `PE-460`, `PE-767`, and `PE-795`

Proposed GitHub Release Notes:

- All bulk file upload workflows will now bundle as many transactions as possible into one or more ANS-104 bundles, each with a max size limit of 500MiB or 500 data items
- Entity naming rules are now being enforced on ArFS write actions:
  - Name must not be empty
  - No null characters
  - Name must be less than 255 bytes
- Uploads now have support for custom content types
- Files uploaded with no extension now inherit the correct content type by default
- Wallet balance checks when using fee boost now assert the correct balance


matibat (approved):
The release notes look good to me :+1:

PR #171, opened 2022-02-28
PE-1058: do not print keys unless explicitly specified

matibat:


src/ardrive.ts
```diff
@@ -913,22 +914,51 @@ export class ArDrive extends ArDriveAnonymous {
...
+					drive.entityType,
+					drive.name,
+					drive.txId,
+					drive.unixTime,
+					drive.drivePrivacy,
+					drive.rootFolderId,
+					drive.driveAuthMode,
+					drive.cipher,
+					drive.cipherIV
+			  );
```
fedellen, line 946: After seeing how this constructor ends up being used, it surprises me that `ArFSPrivateDriveKeyless` doesn't just take a singular drive parameter.

fedellen (approved):
LGTM 👌

src/ardrive.ts
```diff
@@ -913,22 +914,51 @@ export class ArDrive extends ArDriveAnonymous {
...
+		withKeys
+	}: GetPrivateFolderParams): Promise<ArFSPrivateFolder> {
 		if (!owner) {
 			owner = await this.arFsDao.getDriveOwnerForFolderId(folderId);
 		}
 		await this.assertOwnerAddress(owner);
 
-		return this.arFsDao.getPrivateFolder(folderId, driveKey, owner);
+		const folder = await this.arFsDao.getPrivateFolder(folderId, driveKey, owner);
+		return withKeys ? folder : new ArFSPrivateFolderKeyless(folder);
```
karlprieb, line 961: I don't get why `ArFSPrivateFolderKeyless` receives the entity as constructor param and `ArFSPrivateDriveKeyless` just don't.

tests/integration/ardrive.int.test.ts
```diff
@@ -201,6 +201,31 @@ describe('ArDrive class - integrated', () => {
...
+			});
+
+			it('returns the with-keys version of the entitites if withKeys is true', async () => {
+				const drive = await arDrive.getPrivateDrive({
+					driveId: stubEntityID,
+					owner: walletOwner,
+					driveKey: await getStubDriveKey(),
+					withKeys: true
+				});
+				expect(drive.driveKey).to.not.be.undefined;
```
arielmelendez, line 225: Asserting the correct value would be best

arielmelendez (approved):

src/ardrive.ts
karlprieb: Haha I also made a comment about that 🤣

karlprieb (approved):

tests/integration/ardrive.int.test.ts
matibat: [311f15b](https://github.com/ardriveapp/ardrive-core-js/pull/171/commits/311f15b79d2775bcbac705e887efe3b83e409d5e)

src/ardrive.ts
matibat: As the safeDriveBuilder returns a keyless version, it's hard to 1st declare an ArFSPrivateDrive to then wrap it into ArFSPrivateDriveKeyless as there is the case where no keys are provided

PR #174, opened 2022-02-28
PE-943: Quick implementation of a persistent cache

arielmelendez:


src/arfs/arfs_metadata_cache.int.test.ts
```diff
@@ -0,0 +1,42 @@
...
+import path from 'path';
+import * as fs from 'fs';
+import * as os from 'os';
+import { TxID } from '../exports';
+import { v4 } from 'uuid';
+import { ArFSMetadataCache } from '../arfs/arfs_metadata_cache';
+
+describe('ArFSMetadataCache class -- integrated', () => {
+	const homeDir = os.homedir();
+	const metadataCacheDir = path.join(homeDir, '.ardrive', 'caches', 'metadata');
```
karlprieb, line 11: dotfiles are valid on windows after Windows 10 1903 (May, 2019). If we want to support older Windows we will need to rename `.ardrive` folder

src/arfs/arfs_metadata_cache.ts
```diff
@@ -0,0 +1,63 @@
...
+	private static shouldCacheLog = process.env['ARDRIVE_CACHE_LOG'] === '1';
+
+	static async getCacheFolder(): Promise<string> {
+		// Don't kick off another setup while setup is in progress
+		if (this.cacheFolderPromise) {
+			return this.cacheFolderPromise;
+		}
+
+		const homeDir = os.homedir();
+		const metadataCacheDir = path.join(homeDir, '.ardrive', 'caches', 'metadata');
```
karlprieb, line 17: dotfiles are valid on windows after Windows 10 1903 (May, 2019). If we want to support older Windows we will need to rename `.ardrive` folder

matibat: We could have a `getArdriveHome` method and an `ARDRIVE_HOME` env variable. We can decide at `getArdriveHome` if we use `"${HOME}/.ardrive"`, `"${HOME}/ardrive"`, or the one provided by the user.

I'd suggest for `getArdriveHome` to use the following logic:
- Is there a valid directory set at `ARDRIVE_HOME`?
  - then return that
- Is the current platform Windows?
  - then return ``${homeDir}\ardrive``
- else the system support using point
  - return ``${homeDir}/.ardrive``

We can later decide whether or not to check for the Windows version to be more sophisticated

matibat: We can make the cupcake of simply checking for the Windows case

src/arfs/arfs_metadata_cache.int.test.ts
```diff
@@ -0,0 +1,42 @@
...
+	afterEach(() => {
+		// clean up specific cache files
+		if (fs.existsSync(cachedTxFilePath)) {
+			fs.rmSync(cachedTxFilePath);
+		}
+	});
+
+	it('returns expected cached data dependent on cached data availability', async () => {
+		// Use randomized text between tests to ensure that stale data doesn't affect the test
+		const expectedRandomBuffer = Buffer.from(v4());
```
matibat, line 31: We usually avoid using random input data at the unit tests. I guess it's fine for this specific case of cache

src/utils/get_tx_data.ts
```diff
@@ -0,0 +1,38 @@
+import Arweave from 'arweave';
+import axios, { AxiosInstance } from 'axios';
+import axiosRetry, { exponentialDelay } from 'axios-retry';
+import { ArFSMetadataCache } from '../arfs/arfs_metadata_cache';
+import { gatewayUrlForArweave, TransactionID } from '../exports';
+
+export async function getDataForTxID(
+	txId: TransactionID,
+	arweave: Arweave,
+	axiosInstance?: AxiosInstance // Provided for test mocking :(
```
matibat, line 10: It would be a goal :soccer: if we could set up global mocking properly!

matibat: BTW we could have a default here and get rid of the line #19 above (`axiosInstance ??= axios.create();`)
```suggestion
	axiosInstance: AxiosInstance = axios.create() // Provided for test mocking :(
```

matibat (commented):
Just dropped some comments here but looking good overall!

src/arfs/arfs_metadata_cache.int.test.ts
arielmelendez: I agree! Randomness is typically not good, but it's being used here as a deliberate strategy for avoiding false positives caused by parallel execution of tests.

src/utils/get_tx_data.ts
arielmelendez: 2ecc89c

src/arfs/arfs_metadata_cache.ts
arielmelendez: e404a6c

matibat: I'm good with that :+1:

matibat (dismissed):
It looks good to me.
A future improvement here could be to properly manage FS permissions (this code will fail if running as a user with no writing permissions over their home dir).

karlprieb (approved):

src/utils/get_tx_data.ts
```diff
@@ -0,0 +1,37 @@
+import Arweave from 'arweave';
+import axios, { AxiosInstance } from 'axios';
+import axiosRetry, { exponentialDelay } from 'axios-retry';
+import { ArFSMetadataCache } from '../arfs/arfs_metadata_cache';
+import { gatewayUrlForArweave, TransactionID } from '../exports';
+
+export async function getDataForTxID(
+	txId: TransactionID,
+	arweave: Arweave,
```
fedellen, line 9: Instead of passing in a heavy `arweave` instance here we could simply take a `URL` param. The only trade-off would be that the caller would need to use `gatewayUrlForArweave`. That could be explained in a doc-block

fedellen (approved):
LGTM!

karlprieb (approved):
👍

fedellen (approved):

PR #179, opened 2022-03-03
Parellel chunks experimental

fedellen:
This PR implements our own `TransactionUploader` from arweave-js that is capable of uploading chunks in parallel.

This is in progress, I'll mark it as draft for now and update this description when we're further along

### DIFF
--- a/src/arfs/arfs_transaction_uploader.ts
+++ b/src/arfs/arfs_transaction_uploader.ts
@@ -84,9 +84,11 @@ export class ArFSTransactionUploader {
 	}
 
 	/**
+	 * TODO: Update this docblock, and add docs to other methods
+	 *
 	 * Uploads the next part of the transaction.
 	 * On the first call this posts the transaction
-	 * itself and on any subsequent calls uploads the
+	 * itself and on unknown subsequent calls uploads the
 	 * next chunk until it completes.
 	 */
 	public async batchUploadChunks(): Promise<void> {
@@ -95,7 +97,7 @@ export class ArFSTransactionUploader {
 		}
 
 		if (!this.txPosted) {
-			await this.postTransactionHeader();
+			await this.postTransactionHeaders();
 			return;
 		}
 
@@ -119,10 +121,55 @@ export class ArFSTransactionUploader {
 	}
 
 	private async uploadChunk(chunk: Chunk) {
-		let resp: AxiosResponse<any> | string;
+		try {
+			console.log('sending chunk');
+			await this.retryRequestUntilMaxErrors(this.arweave.api.post(`chunk`, chunk));
+		} catch (err) {
+			throw new Error(`Too many errors encountered while posting chunks: ${err}`);
+		}
+
+		this.uploadedChunks++;
+
+		if (this.chunkOffset < this.totalChunks) {
+			console.log('Resident Set Size:', formatBytes(+process.memoryUsage().rss));
+			// Start next chunk when this one finishes
+			console.log('getting new chunk at offset:', this.chunkOffset);
+			await this.uploadChunk(this.transaction.getChunk(this.chunkOffset++, this.transaction.data));
+		}
+		return;
+	}
+
+	// POST to /tx
+	private async postTransactionHeaders(): Promise<void> {
+		const uploadInBody = this.totalChunks <= MAX_CHUNKS_IN_BODY;
+
+		// We will send the data with the headers if chunks will fit into transaction header body
+		// Otherwise we send the headers with no data
+		const transactionToUpload = uploadInBody
+			? this.transaction
+			: new Transaction(Object.assign({}, this.transaction, { data: new Uint8Array(0) }));
 
 		try {
-			resp = await this.arweave.api.post(`chunk`, chunk);
+			console.log('posting tx');
+			await this.retryRequestUntilMaxErrors(this.arweave.api.post(`tx`, transactionToUpload));
+		} catch (err) {
+			throw new Error(`Too many errors encountered while posting transaction headers:  ${err}`);
+		}
+
+		this.txPosted = true;
+
+		if (uploadInBody) {
+			this.chunkOffset += MAX_CHUNKS_IN_BODY;
+			this.uploadedChunks += MAX_CHUNKS_IN_BODY;
+		}
+		return;
+	}
+
+	private async retryRequestUntilMaxErrors(request: Promise<AxiosResponse<unknown>>) {
+		let resp: AxiosResponse<unknown> | string;
+
+		try {
+			resp = await request;
 		} catch (err) {
 			resp = err.message;
 		}
@@ -136,82 +183,28 @@ export class ArFSTransactionUploader {
 				this.totalErrors++;
 				console.log('this.totalErrors', this.totalErrors);
 				if (this.totalErrors >= MAX_ERRORS) {
-					throw new Error(`Unable to complete chunk upload: ${error}`);
+					throw new Error(`Unable to complete request: ${error}`);
 				} else {
 					console.log('delaying');
-					// Jitter delay after failed chunk uploads -- subtract up to 30% from 40 seconds
+					// Jitter delay after failed requests -- subtract up to 30% from ERROR_DELAY
 					await new Promise((res) => setTimeout(res, ERROR_DELAY - ERROR_DELAY * Math.random() * 0.3));
 
-					// Retry the chunk
+					// Retry the request
 					console.log('Resident Set Size:', formatBytes(+process.memoryUsage().rss));
-					console.log('retrying a chunk');
-					await this.uploadChunk(chunk);
+					console.log('retrying request');
+					await request;
 				}
-			}
-		} else {
-			this.uploadedChunks++;
-
-			if (this.chunkOffset < this.totalChunks) {
-				console.log('Resident Set Size:', formatBytes(+process.memoryUsage().rss));
-				// Start next chunk when this one finishes
-				console.log('getting new chunk at offset:', this.chunkOffset);
-				await this.uploadChunk(this.transaction.getChunk(this.chunkOffset++, this.transaction.data));
-			}
-		}
-		return;
-	}
-
-	// POST to /tx
-	private async postTransactionHeader(): Promise<void> {
-		const uploadInBody = this.totalChunks <= MAX_CHUNKS_IN_BODY;
-
-		// We will send the data with the headers if chunks will fit into transaction header body
-		// Otherwise we send the headers with no data
-		const transactionToUpload = uploadInBody
-			? this.transaction
-			: new Transaction(Object.assign({}, this.transaction, { data: new Uint8Array(0) }));
-
-		let resp: AxiosResponse<any> | string;
-
-		try {
-			console.log('posting');
-			resp = await this.arweave.api.post(`tx`, transactionToUpload);
-		} catch (err) {
-			resp = err.message;
-		}
-
-		if (respIsError(resp) || resp.status !== 200) {
-			const error = respIsError(resp) ? resp : getError(resp);
-
-			if (this.totalErrors >= MAX_ERRORS) {
-				throw new Error(`Unable to complete transaction post: ${error}`);
-			} else {
-				console.log('delaying');
-				// Jitter delay after failed chunk uploads -- subtract up to 30% from 40 seconds
-				await new Promise((res) => setTimeout(res, ERROR_DELAY - ERROR_DELAY * Math.random() * 0.3));
-
-				// Retry the chunk
-				console.log('Resident Set Size:', formatBytes(+process.memoryUsage().rss));
-				console.log('retrying post');
-				await this.postTransactionHeader();
-			}
-		} else {
-			this.txPosted = true;
-
-			if (uploadInBody) {
-				this.chunkOffset = MAX_CHUNKS_IN_BODY;
-				this.uploadedChunks += MAX_CHUNKS_IN_BODY;
 			}
 		}
 	}
 }
 
-function respIsError(resp: AxiosResponse<any> | string): resp is string {
+function respIsError(resp: AxiosResponse<unknown> | string): resp is string {
 	return resp === typeof 'string';
 }
 
 // Temp copy pasted from arweave-js for debugging
-export function getError(resp: AxiosResponse<any>) {
+export function getError(resp: AxiosResponse<unknown>) {
 	let data = resp.data;
 
 	if (typeof resp.data === 'string') {
@@ -225,7 +218,7 @@ export function getError(resp: AxiosResponse<any>) {
 	if (resp.data instanceof ArrayBuffer || resp.data instanceof Uint8Array) {
 		console.log('is buffer or such');
 		try {
-			data = JSON.parse(data.toString());
+			data = JSON.parse((data as ArrayBuffer).toString());
 			// eslint-disable-next-line no-empty
 		} catch (e) {}
 	}
@@ -238,5 +231,7 @@ export function getError(resp: AxiosResponse<any>) {
 		writeFileSync(`${Math.random()}.txt`, JSON.stringify(resp, null, 4));
 	}
 
+	// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+	// @ts-ignore
 	return data ? data.error || data : resp.statusText || 'unknown';
 }
--- a/src/arfs/arfs_transaction_uploader.ts
+++ b/src/arfs/arfs_transaction_uploader.ts
@@ -15,7 +15,6 @@ interface Chunk {
 
 // Maximum amount of chunks we will upload in the body.
 const MAX_CHUNKS_IN_BODY = 1;
-const MAX_CHUNKS_BATCH_SIZE = 32;
 
 const MAX_ERRORS = 100;
 
@@ -61,8 +60,17 @@ export class ArFSTransactionUploader {
 
 	private arweave: Arweave;
 	private transaction: Transaction;
+	private maxConcurrentChunks: number;
 
-	constructor({ transaction, arweave }: { transaction: Transaction; arweave: Arweave }) {
+	constructor({
+		transaction,
+		arweave,
+		maxConcurrentChunks = 32
+	}: {
+		transaction: Transaction;
+		arweave: Arweave;
+		maxConcurrentChunks?: number;
+	}) {
 		if (!transaction.id) {
 			throw new Error(`Transaction is not signed`);
 		}
@@ -70,10 +78,9 @@ export class ArFSTransactionUploader {
 			throw new Error(`Transaction chunks not prepared`);
 		}
 
-		// Make a copy of transaction, zeroing the data so we can serialize.
-		// this.data = transaction.data;
 		this.arweave = arweave;
 		this.transaction = transaction;
+		this.maxConcurrentChunks = maxConcurrentChunks;
 	}
 
 	/**
@@ -88,13 +95,13 @@ export class ArFSTransactionUploader {
 		}
 
 		if (!this.txPosted) {
-			await this.postTransaction();
+			await this.postTransactionHeader();
 			return;
 		}
 
 		console.log('this.totalChunks', this.totalChunks);
 		console.log('this.chunkOffset', this.chunkOffset);
-		const initialChunksToGet = Math.min(this.totalChunks - this.chunkOffset, MAX_CHUNKS_BATCH_SIZE);
+		const initialChunksToGet = Math.min(this.totalChunks - this.chunkOffset, this.maxConcurrentChunks);
 		console.log('getting this many chunks', initialChunksToGet);
 
 		const initialChunks = (() => {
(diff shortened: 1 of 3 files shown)


src/types/entity_key.test.ts
```diff
@@ -11,7 +11,7 @@ describe('EntityKey class', () => {
 
 	it('throws if a non-buffer is given', () => {
 		// eslint-disable-next-line prettier/prettier
-		const nonBuffer: Buffer = ('non buffer type' as unknown) as Buffer;
+		const nonBuffer: Buffer = 'non buffer type' as unknown as Buffer;
```
fedellen, line 14: The formatter seems to be ignoring certain `// esling-disable` calls 🤔

src/arfs/arfs_transaction_uploader.ts
```diff
@@ -0,0 +1,181 @@
...
+
+		await Promise.all(uploadPromises);
+	}
+
+	/**
+	 * Iterates through and posts each chunk to the `/chunk`
+	 * endpoint on the provided gateway
+	 *
+	 * @remarks Will continue posting chunks until all chunks have been posted
+	 */
```
fedellen, line 111: Add `@throws` info

Same with postTransaction when that docblock is added

src/arfs/arfs_transaction_uploader.ts
```diff
@@ -0,0 +1,181 @@
...
+
+			if (FATAL_CHUNK_UPLOAD_ERRORS.includes(error)) {
+				throw new Error(`Fatal error uploading chunk ${this.chunkOffset}: ${error}`);
+			} else {
+				this.totalErrors++;
+				if (this.totalErrors >= MAX_ERRORS) {
+					throw new Error(`Unable to complete request: ${error}`);
+				} else {
+					// Jitter delay after failed requests -- subtract up to 30% from ERROR_DELAY
+					await new Promise((res) => setTimeout(res, ERROR_DELAY - ERROR_DELAY * Math.random() * 0.3));
```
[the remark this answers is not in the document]
fedellen: Good call, added here: 831f5b4

src/arfs/arfs_transaction_uploader.ts
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	/**
+	 * Iterates through and posts each chunk to the `/chunk`
+	 * endpoint on the provided gateway
+	 *
+	 * @remarks Will continue posting chunks until all chunks have been posted
+	 */
+	private async uploadChunk(): Promise<void> {
+		while (this.chunkOffset < this.totalChunks) {
```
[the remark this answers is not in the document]
fedellen: Added a `failedRequests` state to handle this here 831f5b4 

Each loop will now skip if there are any failed requests present

src/arfs/arfs_transaction_uploader.ts
```diff
@@ -0,0 +1,181 @@
+import Transaction from 'arweave/node/lib/transaction';
+import axios, { AxiosResponse } from 'axios';
+
+/** Maximum amount of chunks we will upload in the transaction body */
+const MAX_CHUNKS_IN_BODY = 1;
+
+/** Maximum errors we will accumulate before ending the upload attempt */
+const MAX_ERRORS = 100;
```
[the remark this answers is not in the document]
fedellen: Resolved during 831f5b4

src/arfs/arfs_transaction_uploader.ts
```diff
@@ -0,0 +1,208 @@
...
+	'chunk_proof_ratio_not_attractive',
+	'invalid_proof'
+];
+
+interface ArFSTransactionUploaderConstructorParams {
+	gatewayUrl: URL;
+	transaction: Transaction;
+	maxConcurrentChunks?: number;
+	maxRetriesPerRequest?: number;
+}
```
fedellen, line 39: nit: add whitespace

src/arfs/arfs_transaction_uploader.ts
```diff
@@ -0,0 +1,240 @@
...
+ *			transaction,
+ *			gatewayUrl: new URL('https://arweave.net:443')
+ *		});
+ *
+ *		while (!transactionUploader.isComplete) {
+ *			await transactionUploader.batchUploadChunks();
+ *		}
+ * ```
+ */
+export class ArFSTransactionUploader {
```
fedellen, line 60: nit: This class actually has nothing to do with ArFS. Maybe it should be called `ParallelTransactionUploader` or similar?

src/arfs/arfs_transaction_uploader.ts
```diff
@@ -0,0 +1,240 @@
...
+ * 	await transaction.prepareChunks(transaction.data);
+ *
+ *		const transactionUploader = new ArFSTransactionUploader({
+ *			transaction,
+ *			gatewayUrl: new URL('https://arweave.net:443')
+ *		});
+ *
+ *		while (!transactionUploader.isComplete) {
+ *			await transactionUploader.batchUploadChunks();
+ *		}
```
fedellen, line 57: Calling this in a while loop is now unnecessary. This will only ever loop once, uploading the tx header and then the chunks.

Update comment and the implementation in `arfsdao`

src/arfs/arfs_transaction_uploader.ts
```diff
@@ -0,0 +1,238 @@
...
+
+		if (!this.txPosted) {
+			await this.postTransactionHeader();
+
+			if (this.isComplete) {
+				return;
+			}
+		}
+
+		const numOfConcurrentUploadPromises = Math.min(this.totalChunks - this.chunkOffset, this.maxConcurrentChunks);
```
arielmelendez, line 125: Readability suggestion:
```suggestion
		const numRemainingChunks = this.totalChunks - this.chunkOffset;
		const numOfConcurrentUploadPromises = Math.min(numRemainingChunks, this.maxConcurrentChunks);
```

src/arfs/arfs_transaction_uploader.ts
```diff
@@ -0,0 +1,238 @@
...
+	 * @throws when max retries have been exhausted
+	 */
+	private async retryRequestUntilMaxErrors(request: Promise<AxiosResponse<unknown>>) {
+		let resp: AxiosResponse<unknown> | string;
+		let retryNumber = 0;
+		let error = '';
+
+		while (retryNumber <= this.maxRetriesPerRequest && !this.hasFailedRequests) {
+			try {
+				resp = await request;
```
arielmelendez, line 208: How well tested is this? It kinda seems like you're just awaiting the same promise again so you'd just resolve the same error immediately... If I've got that right, then you need to pass in a FUNCTION that returns a promise that you'd then await upon here.

src/arfs/arfs_transaction_uploader.ts
```diff
@@ -0,0 +1,238 @@
...
+
+					retryNumber++;
+				}
+			} else {
+				// Request has succeeded with status code 200, return from loop
+				return;
+			}
+		}
+
+		this.failedRequests.push(request);
```
arielmelendez, line 231: Is it strictly necessary for this to be an array anymore given your implementation? Seems like as soon as one of the chunk requests has failed out, the show is over... Might be able to simply get away with a boolean flag.

src/arfs/arfs_transaction_uploader.ts
```diff
@@ -0,0 +1,238 @@
...
+
+	/**
+	 * Posts the transaction's header to the `/tx` endpoint on the provided gateway
+	 *
+	 * @remarks Will post chunks with header if those chunks will fit into the transaction header's body
+	 *
+	 * @throws when a post header request has exceeded the maxRetries and has failed to post
+	 */
+	private async postTransactionHeader(): Promise<void> {
+		const uploadInBody = this.totalChunks <= MAX_CHUNKS_IN_BODY;
```
arielmelendez, line 170: Make sure to test this functionality carefully. Apparently body data in the tx header request is an antiquated way of sending tx data. But it does still seem to be in arweave-js as written here.

src/arfs/arfs_transaction_uploader.ts
```diff
@@ -0,0 +1,238 @@
...
+		try {
+			await this.retryRequestUntilMaxErrors(axios.post(`${this.gatewayUrl.href}tx`, transactionToUpload));
+		} catch (err) {
+			throw new Error(`Too many errors encountered while posting transaction headers:  ${err}`);
+		}
+
+		this.txPosted = true;
+
+		if (uploadInBody) {
+			this.chunkOffset += MAX_CHUNKS_IN_BODY;
```
arielmelendez, line 187: Nit: I'd increment by totalChunks just to be safe since if MAX_CHUNKS_IN_BODY were greater than 1 and you had fewer chunks than it, these values would then be wrong.

src/arfs/arfs_transaction_uploader.ts
arielmelendez: Maybe `MultiChunkTxUploader`

src/arfs/arfs_transaction_uploader.ts
```diff
@@ -0,0 +1,238 @@
...
+			}
+
+			if (respIsError(resp) || resp?.status !== 200) {
+				error = respIsError(resp) ? resp : resp.statusText;
+
+				if (FATAL_CHUNK_UPLOAD_ERRORS.includes(error)) {
+					throw new Error(`Fatal error uploading chunk ${this.chunkOffset}: ${error}`);
+				} else {
+					// Jitter delay after failed requests
+					const delay = Math.pow(2, retryNumber) * INITIAL_ERROR_DELAY;
```
arielmelendez, line 220: Without any element of randomization, there's not actually any "Jitter" here. Additionally, you should model out how long each retry delay is going to be here and determine whether the range of wait times is reasonable as you approach the max number of retries. Let's model the delay for the first 8 retries here:
Retry wait 1: 500ms
Retry wait 2: 1,000ms
Retry wait 3: 2,000ms
Retry wait 4: 4,000ms
Retry wait 5: 8,000ms
Retry wait 6: 16,000ms
Retry wait 7: 32,000ms
Retry wait 8: 64,000ms

So by the 8th retry, the user will have waited nearly 2 minutes for feedback/error. Many retry libraries give you either or both ways to specify a retry-count dependent retry algorithm and/or a max delay time between retries. You've got the former here, but you might want to consider providing a max delay if we intend to use this beyond 8 retries.

tests/integration/arlocal.int.test.ts
```diff
@@ -317,6 +335,64 @@ describe('ArLocal Integration Tests', function () {
 				expectedTxIdPath: `/${rootFolderTxId}/${rootFolder.txId}/${parentFolder.txId}/${childFolder.txId}/${grandChildFolder.txId}/${fileInGrandChildResult.metadataTxId}`
 			});
 		});
+
+		// This test breaks the arlocal docker image atm and is skipped for now
```
arielmelendez, line 339: Still broken?

arielmelendez (changes requested):
Looks good overall! Some minor things to look into and some naming decisions to make.

tests/integration/arlocal.int.test.ts
fedellen: Yes, it seems to still break as of this morning. I had a few other test cases that I've removed here that also failed. 

This one in particular will succeed when pulling `arlocal` locally and running it with `yarn && yarn start` but it fails within the docker test container. Thats kind of why I left this one in as skipped

Haven't invested much time into the reasons why yet

src/arfs/arfs_transaction_uploader.ts
fedellen: You're right, a boolean is probably fine for now. 

I was hoping to extend some persistence to the failedRequests so that we might be able to resume failed uploads in the future. But that will take a little more consideration.

src/arfs/arfs_transaction_uploader.ts
fedellen: [e21b153](https://github.com/ardriveapp/ardrive-core-js/pull/179/commits/e21b1538c297298b03fdbbcd8695c2178b727776)

src/arfs/arfs_transaction_uploader.ts
fedellen: eee7cd6

src/arfs/multi_chunk_tx_uploader.ts
```diff
@@ -31,7 +31,7 @@ const FATAL_CHUNK_UPLOAD_ERRORS = [
 	'invalid_proof'
 ];
 
-interface ArFSTransactionUploaderConstructorParams {
+interface MultiChunkTxUploaderConstructorParams {
```
karlprieb, line 34: 👍

src/arfs/multi_chunk_tx_uploader.ts
```diff
@@ -47,15 +47,15 @@ interface ArFSTransactionUploaderConstructorParams {
...
+ *		const transactionUploader = new MultiChunkTxUploader({
  *			transaction,
  *			gatewayUrl: new URL('https://arweave.net:443')
  *		});
  *
  *		await transactionUploader.batchUploadChunks();
  * ```
  */
-export class ArFSTransactionUploader {
+export class MultiChunkTxUploader {
```
karlprieb, line 58: 👍

src/arfs/arfs_transaction_uploader.ts
fedellen: [0bcb0e2](https://github.com/ardriveapp/ardrive-core-js/pull/179/commits/0bcb0e257ae8b5e8297b37c62b6dc01c9216c3ed)

karlprieb (dismissed):
Looks great :+1: 

I think we can research about body data in the tx header later as it's being used on `arweave-js` and address that in a new PR if needed.

src/arfs/multi_chunk_tx_uploader.ts
```diff
@@ -0,0 +1,246 @@
...
+ *		});
+ *
+ *		await transactionUploader.batchUploadChunks();
+ * ```
+ */
+export class MultiChunkTxUploader {
+	private chunkOffset = 0;
+	private txPosted = false;
+	private uploadedChunks = 0;
+	public hasFailedRequests = false;
```
arielmelendez, line 62: This allows external actors to change the value. You need a private member and a public getter.

fedellen: Made it private here: 7d75933

We don't need to access this externally

arielmelendez (dismissed):

src/arfs/arfsdao.ts
```diff
@@ -789,18 +798,35 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 		const results: ArFSUploadEntitiesResult = { fileResults: [], folderResults: [], bundleResults: [] };
 		const { fileAndMetaDataPlans, fileDataOnlyPlans, folderMetaDataPlans: folderMetaDataPlan } = v2TxPlans;
 
+		const totalFileAndBundleUploads = fileAndMetaDataPlans.length + fileDataOnlyPlans.length + bundlePlans.length;
+		let uploadsCompleted = 0;
+
+		const logProgress = () => {
+			if (this.shouldProgressLog && totalFileAndBundleUploads > 1) {
+				console.error(
```
karlprieb, line 806: We can use `process.stderr.write` with a string ending with `\r` to move the cursor to the first column every time. With that, we will have the progress log updating the same line.
I think it delivers a better experience, but maybe it's not as useful as you lost the track of the progress.

karlprieb (dismissed):

src/types/entity_key.test.ts
matibat: It usually happens to me when I save and the code is auto-formated. What i generally do is to split the declaration into two steps:
```ts
const nonBufferAsUndefined = 'non buffer type' as unknown;
const nonBuffer: Buffer = nonBufferAsUndefined as Buffer;
```
We might need to take a look at the rules of prettier again :monocle_face:

tests/integration/arlocal.int.test.ts
```diff
@@ -41,48 +41,66 @@ import {
 } from '../../src/arfs/arfs_entities';
 
 describe('ArLocal Integration Tests', function () {
+	this.timeout(60 * 1000); // 60 seconds
```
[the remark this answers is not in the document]
matibat: So is this for a single test? The timeout could be set only at that sope

src/arfs/arfsdao.ts
matibat: That's a good idea! We will need some custom log functions instead of using the `console` methods later on

README.md
```diff
@@ -120,6 +120,27 @@ And finally, to view the detailed error messages in your terminal:
 yarn power-assert -g 'My test case'
 ```
 
+### Progress Logging of Transaction Uploads
```
matibat, line 123: ~Is this section added to the index?~
Nevermind. We don't have one :sweat_smile:

src/arfs/arfsdao.ts
```diff
@@ -214,6 +221,8 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 		super(arweave, undefined, undefined, caches);
 	}
 
+	private shouldProgressLog = process.env['ARDRIVE_PROGRESS_LOG'] === '1';
```
matibat, line 224: Strictly equal to `'1'` VS not empty :opionion:
```suggestion
	private shouldProgressLog = process.env['ARDRIVE_PROGRESS_LOG'] !== '';
```
_Non-blocking comment_

src/arfs/arfsdao.ts
matibat: Actually, it is more complex than that. We log the progress of multiple items, It's hard to display all of them in one single line once a time :opionion:

src/arfs/multi_chunk_tx_uploader.ts
```diff
@@ -0,0 +1,253 @@
...
+		const uploadPromises: Promise<void>[] = [];
+		for (let index = 0; index < numOfConcurrentUploadPromises; index++) {
+			uploadPromises.push(this.uploadChunk());
+		}
+
+		try {
+			await Promise.all(uploadPromises);
+		} catch (err) {
+			throw new Error(err);
+		}
```
matibat, line 138: Does it have different behavior than directly awaiting for `Promise.all`? The error would be thrown anyway...
Unless... the wrapping `Error` at the catch is doing something relevant, in that case:
```suggestion
		await Promise.all(uploadPromises).catch((err) => {
			throw new Error(err);
		})
```

src/arfs/multi_chunk_tx_uploader.ts
```diff
@@ -0,0 +1,253 @@
...
+	 */
+	private async uploadChunk(): Promise<void> {
+		while (this.chunkOffset < this.totalChunks && !this.hasFailedRequests) {
+			const chunk = this.transaction.getChunk(this.chunkOffset++, this.transaction.data);
+
+			try {
+				await this.retryRequestUntilMaxErrors(() => axios.post(`${this.gatewayUrl.href}chunk`, chunk));
+			} catch (err) {
+				throw new Error(`Too many errors encountered while posting chunks: ${err}`);
+			}
```
matibat, line 157: Same comment than above :opionion:

src/arfs/multi_chunk_tx_uploader.ts
```diff
@@ -0,0 +1,253 @@
...
+		// Otherwise we send the headers with no data
+		const transactionToUpload = uploadInBody
+			? this.transaction
+			: new Transaction(Object.assign({}, this.transaction, { data: new Uint8Array(0) }));
+
+		try {
+			await this.retryRequestUntilMaxErrors(() => axios.post(`${this.gatewayUrl.href}tx`, transactionToUpload));
+		} catch (err) {
+			throw new Error(`Too many errors encountered while posting transaction headers:  ${err}`);
+		}
```
matibat, line 186: Same comment than above :opionion:

tests/integration/arlocal.int.test.ts
```diff
@@ -317,6 +335,64 @@ describe('ArLocal Integration Tests', function () {
...
+						destFolderId: rootFolderId,
+						wrappedEntity: wrapFileOrFolder('tests/stub_files/5MiB.txt'),
+						destName: 'unique_0'
+					}
+				]
+			});
+			await arweave.api.get(`mine`);
+
+			// Wait for 10 second for arlocal gateway to process bundle ??
+			await sleep(10000);
```
matibat, line 353: There's no way of directly reaching the ArLocal lib to check for the bundle status?

matibat (approved):
Some non-blocking comments here. Looks good to me overall :sunglasses:

src/arfs/arfsdao.ts
karlprieb: Yep.. it's complex if we log more than one upload at a time.

tests/integration/arlocal.int.test.ts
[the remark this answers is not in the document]
fedellen: Removed here in test file cleanup: cd2645f

tests/integration/arlocal.int.test.ts
fedellen: Removed here in test file cleanup: cd2645f

src/arfs/multi_chunk_tx_uploader.ts
fedellen: Removed this one: [4de6ca5](https://github.com/ardriveapp/ardrive-core-js/pull/179/commits/4de6ca56e6acdce8133df6d8b696e1e24b22ba0a)

src/arfs/multi_chunk_tx_uploader.ts
fedellen: I think I'll leave this two so we know which part of the upload errored out

src/arfs/arfsdao.ts
arielmelendez: This is following the same pattern we used for cache logging. I don't like the "not empty" check because it means the following might not work as expected:
```
export ARDRIVE_PROGRESS_LOG=false
export ARDRIVE_PROGRESS_LOG=0
export ARDRIVE_PROGRESS_LOG=no
etc.
```

src/arfs/arfsdao.ts
arielmelendez: This is a utility feature at the moment and will get replaced with a proper logging system later on. I wouldn't overthink this for now.

arielmelendez (dismissed):

src/arfs/multi_chunk_tx_uploader.ts
matibat: > However, I would advise against ever mixing async/await syntax with then/catch syntax.

Hmmm, are there any disadvantages by doing that? I find it completely safe and readable :opionion:

src/arfs/multi_chunk_tx_uploader.ts
matibat: I'd rather prefer using then..catch instead of the try..catch with promises. But that's only code style, both are equivalent :+1:

matibat (approved):
Looks good to me

src/arfs/multi_chunk_tx_uploader.ts
fedellen: Just a syntax nit. They both do the same thing. 

With this example of simply adding `.catch` to awaited promises, it is hard to even argue that it would be confusing. I think `await Promise.catch((e) => throw(e);` would be okay. I'd still learn towards just wrapping it in a try catch for quicker readability on the indentation and spacing.

Mainly I'd just avoid mixing `.then` and `await`, because that one can lead to confusion and defeat the purpose of async/await syntax.

Either way, just a code style opinion.

src/arfs/arfsdao.ts
```diff
@@ -1064,14 +1098,35 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
+					gatewayApi: new GatewayAPI({ gatewayUrl: gatewayUrlForArweave(this.arweave) }),
+					progressCallback: shouldProgressLog
+						? (pct: number) => {
+								if (!progressLogDebounce || pct === 100) {
+									console.error(`Transaction ${transaction.id} Upload Progress: ${pct}%`);
+									progressLogDebounce = true;
+
+									setTimeout(() => {
+										progressLogDebounce = false;
+									}, 500); // .5 sec debounce
```
matibat, line 1122: If we ever need to use a debouncer at any other place, It would be excellent to have a class for it

matibat (approved):
Nice!

PR #183, opened 2022-03-11
PE-1104: Multi chunk tx uploader unit tests

fedellen:
This PR adds unit test coverage for the MultiChunkTxUploader class.

There was one minor error discovered during test creation: `respIsError` was not properly type checking for string. This is a non-risk change as it only modifies the information being returned from an error and does not effect the behavior of any flow.

### DIFF
--- a/src/arfs/multi_chunk_tx_uploader.test.ts
+++ b/src/arfs/multi_chunk_tx_uploader.test.ts
@@ -26,7 +26,10 @@ describe('MultiChunkTxUploader class', function () {
 	before(async () => {
 		wallet = await arweave.wallets.generate();
 
+		/** 1 chunk of data */
 		const smallData = new Uint8Array(5);
+
+		/** 20 chunks of data */
 		const largeData = new Uint8Array(5_000_000);
 
 		const prepTx = async (data: Uint8Array) => {
@@ -46,15 +49,12 @@ describe('MultiChunkTxUploader class', function () {
 	const progressCallback: ProgressCallback = (pct) => {
 		progressCount++;
 
-		// With exactly 20 chunks sent, we expect 5% progress on each progress callback
+		// With exactly 20 chunks sent, we will expect 5% progress on each progress callback
 		expect(pct).to.equal(progressCount * 5);
-	};
-	const resetProgressCount = () => {
-		progressCount = 0;
 	};
 
 	beforeEach(() => {
-		resetProgressCount();
+		progressCount = 0;
 	});
 
 	const getTxUploader = (transaction: Transaction, maxConcurrentChunks?: number, maxRetriesPerRequest?: number) =>
--- a/src/arfs/multi_chunk_tx_uploader.test.ts
+++ b/src/arfs/multi_chunk_tx_uploader.test.ts
@@ -1,0 +1,335 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import Arweave from 'arweave';
+import Transaction from 'arweave/node/lib/transaction';
+import { JWKInterface } from 'arweave/node/lib/wallet';
+import { AxiosResponse } from 'axios';
+import { expect } from 'chai';
+import { describe } from 'mocha';
+import { spy, stub } from 'sinon';
+import { expectAsyncErrorThrow } from '../../tests/test_helpers';
+import { FATAL_CHUNK_UPLOAD_ERRORS, MultiChunkTxUploader, ProgressCallback } from './multi_chunk_tx_uploader';
+
+describe('MultiChunkTxUploader class', function () {
+	const arweave = Arweave.init({
+		host: 'fake',
+		port: 433,
+		protocol: 'http',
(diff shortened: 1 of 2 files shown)


src/arfs/multi_chunk_tx_uploader.test.ts
```diff
@@ -0,0 +1,335 @@
...
+
+	describe('batchUploadChunks public method', () => {
+		it('throws an error if uploader already has failed requests', async () => {
+			const uploader = getSmallTxUploader();
+
+			const uploadChunkSpy = spy(uploader as any, 'uploadChunk');
+			const postTxHeaderSpy = spy(uploader as any, 'postTransactionHeader');
+			const retryRequestSpy = spy(uploader as any, 'retryRequestUntilMaxErrors');
+
+			uploader['hasFailedRequests'] = true;
```
arielmelendez, line 81: I'm worried about this approach to validation. It's too interested in internal (PRIVATE!) state and not interested in the external presentation of the data (which is what matters most). You really shouldn't be fighting typescript this way, otherwise it means you've got a problem with your type design. Made a readonly getter if you really need to inspect this externally.

src/arfs/multi_chunk_tx_uploader.test.ts
```diff
@@ -0,0 +1,335 @@
...
+			expect(postTxHeaderSpy.calledOnce).to.be.true;
+		});
+	});
+
+	describe('uploadChunk private method', () => {
+		it('sends one chunk and increments offset by one on an unawaited call', () => {
+			const uploader = getLargeTxUploader();
+			const retryRequestSpy = spy(uploader as any, 'retryRequestUntilMaxErrors');
+
+			uploader['uploadChunk']();
```
arielmelendez, line 149: I'm feeling very conflicted about this approach 😬

src/arfs/multi_chunk_tx_uploader.test.ts
```diff
@@ -0,0 +1,335 @@
...
+		const successfulRequestPromise = Promise.resolve(({
+			statusCode: 200
+			// eslint-disable-next-line prettier/prettier
+		} as unknown) as AxiosResponse<unknown>);
+
+		it('throws an error after maxRetries have been exhausted', async function () {
+			const txUploader = getSmallTxUploader(2);
+
+			// With uploader set to 2 retries, we will expect 3,500ms of retries
+			this.timeout(4_200);
```
arielmelendez, line 279: Why the 700ms spread between the test timeout and the expected amount of retry time? This seems brittle and prone to failure when run in different environments.

fedellen: These two tests with the delay timeout seem to take between 3550ms and 3750ms on my end. I just gave it some extra breathing room here. It could be reduced to 3800 on my environment

src/arfs/multi_chunk_tx_uploader.test.ts
fedellen: Theres no need to inspect the private state externally. We're setting this directly here in order to get test coverage on an uploader being called that already has existing failed requests.

I don't see anything wrong with accessing/modifying/stubbing private state/methods during a unit test to gain coverage over parts of the logic. In the end its all JavaScript, private effectively means nothing unless we start using 
actual private members with `#` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields)

Ideally, all of these edge cases are testable via the public methods and members. But at this point that would be a larger re-work to a ticket already being QA'd. Just trying to get some test coverage over these harder to test flows from `TransactionUploader`. We can certainly skip these tests, ArLocal is running them through integration

src/arfs/multi_chunk_tx_uploader.test.ts
fedellen: Changed this to have less retries. And it seems that when the uploader was set to 1 retry it would actually retry twice here. So I've changed that to retry the expected amount of times: db24b56

arielmelendez: Hmm. Your change seems to make the API conflate tries vs. retries. `<=` represents "retries" (as the name implies) while `<` represents "tries".

src/arfs/multi_chunk_tx_uploader.test.ts
arielmelendez: My statement about readonly getters was off base in this case, but everything else applies. I think this is fine for now, but I don't think it's a tool we should reach for often.

matibat (approved):
Looks good to me

fedellen:
Marking as draft for now.

Most of the test coverage here was moved to the GatewayAPI class here: https://github.com/ardriveapp/ardrive-core-js/pull/184

Will reevaluate if we need any of this once its reintegrated.

fedellen:
Closing, PR moved to here: https://github.com/ardriveapp/ardrive-core-js/pull/185

PR #184, opened 2022-03-14
PE-1112: Gateway API refactor

fedellen:
This PR extracts the gateway retry logic into a central class that can be re-used for retrying all requests.

### DIFF
--- a/src/utils/gateway_api.ts
+++ b/src/utils/gateway_api.ts
@@ -1,30 +1,40 @@
 import Transaction from 'arweave/node/lib/transaction';
-import axios, { AxiosResponse } from 'axios';
+import axios, { AxiosInstance, AxiosResponse } from 'axios';
 import { Chunk } from '../arfs/multi_chunk_tx_uploader';
 import { FATAL_CHUNK_UPLOAD_ERRORS, INITIAL_ERROR_DELAY } from './constants';
 
 interface GatewayAPIConstParams {
 	gatewayUrl: URL;
 	maxRetriesPerRequest?: number;
-	/** MS to wait after first failed request; uses exponential delay on subsequent retries */
 	initialErrorDelayMS?: number;
-	/** Errors returned by response.statusText that we should never continue on */
 	fatalErrors?: string[];
-	/** Status codes returned by response.status that we will consider a successful response */
 	validStatusCodes?: number[];
+	axiosInstance?: AxiosInstance;
 }
 
-// With the current default error delay and max retries, we expect the following wait times:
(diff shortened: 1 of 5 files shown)


src/utils/gateway_api.ts
```diff
@@ -0,0 +1,140 @@
...
+
+	private isRequestSuccessful(): boolean {
+		return this.validStatusCodes.includes(this.lastRespStatus);
+	}
+
+	private throwIfFatalError() {
+		if (this.fatalErrors.includes(this.lastError)) {
+			throw new Error(`Fatal error encountered: (Status: ${this.lastRespStatus}) ${this.lastError}`);
+		}
+	}
```
matibat, line 134: These methods check for variables set at the global scope of the class; my guess is that when there's someone using the same instance within multiple promises it will have undefined behavior. Maybe I'm wrong as JavaScript is single-threaded, but there might be an edge case :thinking:

matibat (commented):
Left a comment about one single concern regarding async operations with the same instance of `GatewayAPI`

src/utils/gateway_api.ts
fedellen: Everything after the awaited request that depends on these `lastRespStatus` and `lastError` class members is synchronous. In the single threaded environment we should be absolutely fine there.

I can see the concern though. Perhaps it is best to create a `GatewayAPI` instance on a per-request basis. They are fairly lightweight

src/utils/gateway_api.ts
```diff
@@ -0,0 +1,140 @@
...
+	 *
+	 * @throws when a fatal error has been returned by request
+	 * @throws when max retries have been exhausted
+	 */
+	private async retryRequestUntilMaxRetries(
+		request: () => Promise<AxiosResponse<unknown>>
+	): Promise<AxiosResponse<unknown>> {
+		let retryNumber = 0;
+
+		while (retryNumber <= this.maxRetriesPerRequest) {
```
karlprieb, line 92: Why not:
```suggestion
		while (retryNumber < this.maxRetriesPerRequest) {
```
Using `<=` will make one more retry that `maxRetriesPerRequest`. I think when one set `maxRetriesPerRequest` to 8, one expect to have 8 retries and not 9.

src/utils/gateway_api.ts
matibat: I was rubber-duck debugging the code and didn't even find a way to test this case. I'd say we are completely safe :duck:

karlprieb (approved):

PR #192, opened 2022-04-30
feat(json fields): Expose sourceUri, entityName, and bundledIn fields on ArDrive JSON output PE-1399

fedellen:


fedellen:
Marking this as draft. It exposes all of the fields but I still need to add tests for all of the new API here

src/types/ardrive_types.ts
```diff
@@ -23,8 +23,11 @@ export type ArFSEntityDataType = 'drive' | 'folder' | 'file' | 'bundle';
 
 export interface ArFSEntityData {
 	type: ArFSEntityDataType;
+	sourceUri?: string;
```
arielmelendez, line 26: Can this use the SourceUri type?

src/types/types.ts
```diff
@@ -21,6 +21,8 @@ export type FileKey = EntityKey;
 
 export type DataContentType = string;
 
+export type SourceUri = string;
```
arielmelendez, line 24: I'm starting to wonder if I've got buyer's remorse on the choice of Uri over Url since the former generally requires a scheme (e.g. file:// or something like that. see https://en.wikipedia.org/wiki/Uniform_Resource_Identifier and https://stackoverflow.com/questions/44692872/is-a-local-file-path-a-uri). I just doubt we'll want to absorb the pain right now of mapping file paths to file Uris (if we want to be pedantic).

src/arfs/arfs_file_wrapper.ts
```diff
@@ -104,6 +104,10 @@ export class ArFSManifestToUpload extends ArFSDataToUpload {
 	manifest: Manifest;
 	lastModifiedDateMS: UnixTime;
 
+	public get sourceUri(): SourceUri {
+		return this.destinationBaseName;
```
arielmelendez, line 108: What's the rationale here?

src/arfs/arfs_file_wrapper.ts
```diff
@@ -217,6 +221,10 @@ export class ArFSFileToUpload extends ArFSDataToUpload {
 		}
 	}
 
+	public get sourceUri(): SourceUri {
+		return this.filePath;
```
arielmelendez, line 225: Nit: we may need to add `file://` if it's truly a local file path if we want this to be a Uri.

src/arfs/arfs_file_wrapper.ts
```diff
@@ -269,7 +277,11 @@ export class ArFSFolderToUpload extends ArFSBaseEntityToUpload {
 
 	readonly entityType = 'folder';
 
-	constructor(public readonly filePath: FilePath, public readonly fileStats: Stats) {
+	public get sourceUri(): SourceUri {
+		return this.filePath;
```
arielmelendez, line 281: Same nit about `file://`

src/arfs/arfs_entity_result_factory.ts
```diff
@@ -23,16 +33,24 @@ export interface ArFSUploadEntitiesResult {
...
 	folderMetaDataReward?: Winston;
 	driveKey?: DriveKey;
 }
 
-export interface FileResult {
+export interface FileResult extends BaseArFSUploadResult {
 	fileDataTxId: TransactionID;
 	metaDataTxId: TransactionID;
+	sourceUri: string;
+	fileName: string;
```
arielmelendez, line 53: What's the benefit of specializing to fileName/folderName here if it's all going to get normalized down to entityName in the end?

src/types/ardrive_types.ts
fedellen: ee49275

src/arfs/arfs_file_wrapper.ts
fedellen: Manifest has no source on disk, so I just used the intended name for the source uri. Maybe it should just return an empty string or throw?

src/arfs/arfs_file_wrapper.ts
fedellen: Does that syntax work on a local file system if I use commands on it like a file path? Trying to make it comfortable for a dev to use these fields directly. I can't seem to use `mv` with that in front. I'll need to do a quick research on that

src/arfs/arfs_entity_result_factory.ts
fedellen: Its only to keep the API in ArFSDAO hard to use incorrectly. Since they're both strings it probably isn't necessary here.

I see a sourceUri with the wrong type there, I'll fix that with this

fedellen: [9ed54e2](https://github.com/ardriveapp/ardrive-core-js/pull/192/commits/9ed54e24fa7923e9e6517dccfd4c167569ea29cd)

src/arfs/arfsdao.ts
```diff
@@ -939,6 +961,17 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 			await this.sendTransactionsAsChunks([bundleTx]);
 
 			uploadsCompleted++;
+
+			results.fileResults.push(
+				...currentBundleResults.fileResults.map((r) => {
```
arielmelendez, line 966: Any concerns about the scalability of these spread operators with large numbers of results?

src/arfs/arfs_file_wrapper.ts
arielmelendez: No it probably would not. The idea behind sourceUri was to normalize the potential data inputs for uploads. `filePath`, for example, wouldn't make much sense for a URL. The `ArFSFileToUpload` might have to evolve into specializations like `ArFSLocalFileToUpload` vs. `ArFSRemoteFileToUpload`.

src/arfs/arfs_file_wrapper.ts
arielmelendez: Does `ArFSBaseEntityToUpload` need to be the one to provide sourceUri or can we push it further down the hierarchy to, say, ArFSFileToUpload?

src/arfs/arfs_entity_result_factory.ts
```diff
@@ -23,17 +35,22 @@ export interface ArFSUploadEntitiesResult {
 	bundleResults: BundleResult[];
 }
 
-export interface FolderResult {
+export interface BaseArFSUploadResult {
+	entityName: EntityName;
+	bundledIn?: TransactionID;
+	sourceUri: SourceUri;
```
arielmelendez, line 41: Might need to make this optional if it just doesn't really apply for manifests (unless we start to accept input manifest files).

fedellen: You're right. We're using the same methods to upload manifests, so it needs to be represented as optional. Especially as we move towards uploading from URL. 

399ab75

src/arfs/arfsdao.ts
fedellen: Possibly. Its mapping objects so there are probably unnecessary references being created. I'll switch up approaches

fedellen: 93c095b

src/arfs/arfs_file_wrapper.ts
fedellen: I tried this out, but we use FileToUpload and ManifestToUpload in the same upload code. I think theres a refactor in there where I can handle the uploads with generics and then use a separate interface for `ArFSLocalEntityToUpload`. But that was turning complex when I started it

src/types/types.ts
fedellen: Hmm. My concern here is that we want fields that are useful for devs. The `filePath` in the local file upload use case is certainly useful. Whereas the source URL for the remote file upload use case is what would be useful. 

I personally don't know the use cases for `file://`, although I'm sure there are plenty. What if we changed the field it up to something like `sourcePath`, then we wouldn't be "wrong" with our implementation

fedellen: Updated: [b2c3a44](https://github.com/ardriveapp/ardrive-core-js/pull/192/commits/b2c3a447a5ceb1af05f7b6240d3ed091c1a96212)

src/arfs/arfs_file_wrapper.ts
fedellen: [b2c3a44](https://github.com/ardriveapp/ardrive-core-js/pull/192/commits/b2c3a447a5ceb1af05f7b6240d3ed091c1a96212)

src/arfs/arfs_file_wrapper.ts
fedellen: [b2c3a44](https://github.com/ardriveapp/ardrive-core-js/pull/192/commits/b2c3a447a5ceb1af05f7b6240d3ed091c1a96212)

src/ardrive.ts
```diff
@@ -864,6 +886,9 @@ export class ArDrive extends ArDriveAnonymous {
 				type: 'bundle',
 				bundleTxId: createDriveResult.bundleTxId
 			});
+			arFSResults.created[0].bundledIn = createDriveResult.bundleTxId;
```
arielmelendez, line 889: Let's iterate over the entries and add the bundledIn field BEFORE pushing the bundle entry itself just to avoid manual indexing.

src/arfs/arfs_file_wrapper.ts
```diff
@@ -18,13 +18,13 @@ import { encryptedDataSize, extToMime } from '../utils/common';
...
 import { ArFSPrivateFile, ArFSPublicFile, ArFSWithPath } from './arfs_entities';
-import { ArFSPublicFileWithPaths, ArFSPublicFolderWithPaths } from '../exports';
+import { ArFSPublicFileWithPaths, ArFSPublicFolderWithPaths, SourceUri } from '../exports';
 import { defaultArweaveGatewayPath } from '../utils/constants';
 
 const pipelinePromise = promisify(pipeline);
 
 type BaseName = string;
-type FilePath = string;
+type EntityPath = string;
```
arielmelendez, line 27: Consider a name that doesn't conflate a local filesystem construct with the (already overloaded) concept of Entities for ArFS.

src/arfs/arfs_file_wrapper.ts
```diff
@@ -76,7 +80,8 @@ export function isFolder(fileOrFolder: ArFSDataToUpload | ArFSFolderToUpload): f
 
 export abstract class ArFSBaseEntityToUpload {
 	abstract getBaseName(): BaseName;
-	abstract entityType: EntityType;
+	abstract readonly entityType: EntityType;
+	readonly sourceUri?: SourceUri;
```
arielmelendez, line 84: Could be worth a comment as to why this is a) not abstract and b) optional.

src/arfs/arfsdao.ts
```diff
@@ -875,6 +884,11 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 			// The upload planner has planned to upload bundles, proceed with bundling
 			let dataItems: DataItem[] = [];
 
+			const currentBundleResults: { folderResults: FolderResult[]; fileResults: FileResult[] } = {
```
arielmelendez, line 887: Consider a comment to explain the presence of this structure.

arielmelendez (approved):

src/arfs/arfsdao.ts
fedellen: 2984680

src/arfs/arfs_file_wrapper.ts
fedellen: 2984680

arielmelendez (approved):

PR #199, opened 2022-06-01
PE-1651: GQL robustness

fedellen:
This PR includes:

- GQL query post capability with the existing GatewayAPI retry strategy
- `getTxData` function has been migrated to excise the ArFS builder dependency on `arweave-js` and to centralize all retry logic (`axios-retry` dependency was removed with this)
- Logging improvement on all retry strategies
- Improvement on tag filtering performance and spread operator usage for List Public/Private Folder


src/arfs/arfsdao_anonymous.ts
```diff
@@ -253,20 +262,19 @@ export class ArFSDAOAnonymous extends ArFSDAOType {
...
 			const { edges } = transactions;
 			hasNextPage = transactions.pageInfo.hasNextPage;
 			const files: Promise<ArFSPublicFile>[] = edges.map(async (edge: GQLEdgeInterface) => {
 				const { node } = edge;
 				cursor = edge.cursor;
-				const fileBuilder = ArFSPublicFileBuilder.fromArweaveNode(node, this.arweave);
+				const fileBuilder = ArFSPublicFileBuilder.fromArweaveNode(node, this.gatewayApi);
 				const file = await fileBuilder.build(node);
 				const cacheKey = { fileId: file.fileId, owner };
+				allFiles.push(file);
```
karlprieb, line 274: do we need `allFiles`? seems that `files` and `allFiles` are the same

karlprieb (dismissed):
LGTM, I just have some questions, but I don't think they block this PR.

karlprieb (approved):

PR #201, opened 2022-06-06
PE-1682: Release ArDrive Core 1.15.0

fedellen:
Proposed Release Notes:

- GraphQL requests are now more robust and will use a retry strategy
- Additional `stderr` logging added to all retried requests to an Arweave gateway


thiagocarvalhodev (approved):
Approved as it was already approved in dev

PR #203, opened 2022-06-08
PE-1681: sync dev and version bump

fedellen:


matibat (approved):
This is because we merged straight to master. LGTM!

PR #205, opened 2022-07-02
feat(test wallet): Adds testWallet function to WalletDAO PE-1809

arielmelendez:


src/wallet_dao.ts
```diff
@@ -144,4 +144,21 @@ export class WalletDAO {
 			throw new Error(`Transaction failed. Response: ${response}`);
 		}
 	}
+
+	// Attempts to sign, without submitting to Arweave, a stubbed out transaction with the provided wallet
+	async testWallet(wallet: Wallet): Promise<void> {
```
fedellen, line 149: I'm not sure that `testWallet` is a clear enough name here. When I read the method I first assume that I will be using this to create a test wallet for development purposes

Consider `isValidWallet` or `debugWallet` or similar

fedellen, line 149:
```suggestion
	public async testWallet(wallet: Wallet): Promise<void> {
```

fedellen (commented):
Looks good! We should absolutely include this functionality as part of our wallet utils

Left a comment on the method name, everything else is peachy 🍑

matibat (commented):
The CORE-side is looking good. I'm just wondering why the CLI-side PR is closed

matibat:
We are gonna hang on this for now. It might be addressed in the future, but not for now.

PR #207, opened 2022-07-25
PE-1534: custom metadata (CORE only)

matibat:
Changes:

- Follows up the work of Custom Metadata by implementing the WRITE-only side at CORE

Note: Read-only and CLI side will be implemented in another Story

### DIFF
--- a/src/types/custom_metadata_types.ts
+++ b/src/types/custom_metadata_types.ts
@@ -6,14 +6,16 @@ const invalidSchemaErrorMessage = `Invalid custom metadata schema. Please submit
 const customMetaDataGqlTagShapeOne = '{ "TAG_NAME": "TAG_VALUE" }';
 const customMetaDataGqlTagShapeTwo = '{ "TAG_NAME": ["VAL 1", "VAL 2" ] }';
 const customMetaDataJsonShape = '{ "TAG_NAME": { "Any": [ "Valid", "JSON" ] } }';
-const customMetaDataShape = `{ metaDataJson?: ${customMetaDataGqlTagShapeOne}, metaDataGql?: ${customMetaDataGqlTagShapeTwo} }`;
+const customMetaDataShape = `{ metaDataJson?: ${customMetaDataGqlTagShapeOne}, metaDataGql?: ${customMetaDataGqlTagShapeTwo}, dataGqlTags?: ${customMetaDataGqlTagShapeTwo} }`;
 
 export const invalidCustomMetaDataGqlTagErrorMessage = `${invalidSchemaErrorMessage}${customMetaDataGqlTagShapeOne} or ${customMetaDataGqlTagShapeTwo}`;
+export const invalidCustomDataGqlTagErrorMessage = `${invalidSchemaErrorMessage}${customMetaDataGqlTagShapeOne} or ${customMetaDataGqlTagShapeTwo}`;
 export const invalidCustomMetaDataJsonErrorMessage = `${invalidSchemaErrorMessage}${customMetaDataJsonShape}`;
 export const invalidCustomMetaDataErrorMessage = `${invalidSchemaErrorMessage}${customMetaDataShape}`;
 
 export type CustomMetaDataGqlTags = Record<string, string | string[]>;
 export type CustomMetaDataJsonFields = EntityMetaDataTransactionData;
+export type CustomMetaDataTagInterface = CustomMetaDataGqlTags;
 
 export interface CustomMetaData {
 	/** Include custom metadata on MetaData Tx Data JSON */
@@ -24,7 +26,7 @@ export interface CustomMetaData {
 
 	// TODO: Include dataTx GQL tags as an option (PE-1534)
 	/** Include custom metadata on File Data Tx GQL Tags */
-	// dataGqlTags?: CustomMetaDataTagInterface;
+	dataGqlTags?: CustomMetaDataTagInterface;
 }
 
 export function isCustomMetaDataJsonFields(customDataJson: unknown): customDataJson is CustomMetaDataJsonFields {
@@ -91,10 +93,14 @@ export function isCustomMetaData(tags: unknown): tags is CustomMetaData {
 	}
 
 	for (const [key, metaData] of Object.entries(tags)) {
+		// consider using a sqwitch..case
 		if (key === 'metaDataJson' && !isCustomMetaDataJsonFields(metaData)) {
 			return false;
 		}
 		if (key === 'metaDataGqlTags' && !isCustomMetaDataGqlTags(metaData)) {
+			return false;
+		}
+		if (key === 'dataGqlTags' && !isCustomMetaDataGqlTags(metaData)) {
 			return false;
 		}
 	}
@@ -117,6 +123,7 @@ export function assertCustomMetaDataJsonFields(tags: unknown): tags is CustomMet
 
 export function assertCustomMetaData(tags: unknown): tags is CustomMetaData {
 	if (!isCustomMetaData(tags)) {
+		// TODO: throw the error for data ones as well.
 		throw Error(invalidCustomMetaDataErrorMessage);
 	}
 	return true;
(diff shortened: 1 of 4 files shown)


src/types/custom_metadata_types.ts
```diff
@@ -24,7 +26,7 @@ export interface CustomMetaData {
 
 	// TODO: Include dataTx GQL tags as an option (PE-1534)
```
fedellen, line 27: We can remove this todo (:

tests/integration/arlocal.int.test.ts
```diff
@@ -486,6 +486,37 @@ describe('ArLocal Integration Tests', function () {
 			});
 		});
 
+		it('upload a public file as a v2 transaction with custom data gql tags', async () => {
```
fedellen, line 489: This can be re-used for a private test since we're offering that here as well

src/types/custom_metadata_types.ts
matibat: 7d4e9782ff04612984813f8360cd2b53f1f347c4

tests/integration/arlocal.int.test.ts
matibat: I may ask you how to work with private files with the wrapped entities to upload by tomorrow because I've just tried and didn't get how to 😅

src/types/custom_metadata_types.ts
```diff
@@ -91,12 +92,16 @@ export function isCustomMetaData(tags: unknown): tags is CustomMetaData {
 	}
 
 	for (const [key, metaData] of Object.entries(tags)) {
+		// consider using a sqwitch..case
```
arielmelendez, line 95: I've heard of a switch... but where do you get one of these newfangled SQWITCHes?!?

arielmelendez, line 95:
```suggestion
		switch (key) {
			case 'metaDataJson':
			case 'metaDataGqlTags':
			case 'dataGqlTags':
				if (!isCustomMetaDataJsonFields(metaData)) {
					return false;
				}
				break;
			default:
				break;
		}
```
I think it's basically the same number of lines but much clearer.

src/types/custom_metadata_types.ts
```diff
@@ -117,6 +122,7 @@ export function assertCustomMetaDataJsonFields(tags: unknown): tags is CustomMet
 
 export function assertCustomMetaData(tags: unknown): tags is CustomMetaData {
 	if (!isCustomMetaData(tags)) {
+		// TODO: throw the error for data ones as well.
```
arielmelendez, line 125: Is this still TODO?

tests/helpers/arlocal_test_assertions.ts
```diff
@@ -30,6 +30,8 @@ import {
 } from '../../src/types';
 import { getDecodedTags } from '../test_helpers';
 
+const defaultDataContentType = 'text/plain';
```
arielmelendez, line 33: In what sense is this default? Where is that established?

tests/helpers/arlocal_test_assertions.ts
```diff
@@ -350,14 +352,18 @@ export function assertFolderMetaDataGqlTags(
...
-		contentType: DataContentType;
+		contentType?: DataContentType;
+		customMetaData?: CustomMetaDataGqlTags;
 	}
 ): void {
-	const { contentType } = expectations;
+	const { contentType, customMetaData } = expectations;
 	const dataTxTags = getDecodedTags(dataTx.tags);
 
+	const expectedData: GQLTagInterface[] = mapMetaDataTagInterfaceToGqlTagInterface(customMetaData);
```
arielmelendez, line 362: Consider name `expectedCustomTags`

tests/integration/arlocal.int.test.ts
arielmelendez: As much as I'd like to see a change in the naming pattern for the tests in this file, you might want to consider being consistent with it, i.e.
```
		it('we can upload a public file as a v2 transaction with custom data gql tags', async () => {
```

arielmelendez (approved):
A few questions and nits but overall seems like it'll work :)

tests/helpers/arlocal_test_assertions.ts
fedellen: I picked this because most or all of our stub files are txt files.

Although I prefer Karl's answer 🧠

src/types/custom_metadata_types.ts
matibat: It's like a SWITCH but way more cool. That's a feature only available for cool devs 😎

matibat: c771e0b925116bf501f6db0b6d97b25ce3fe4ede

matibat: Nit: I didn't add a default there, should I?

src/types/custom_metadata_types.ts
matibat: Yes... I don't know in that context if they are parsing the dataTx or metaDataTx. That's only to print the correct error message, e.i. "error in your dataTx tags" instead of "error in your metaDataTx tags"

tests/helpers/arlocal_test_assertions.ts
matibat: b8ca31d1c9cb9b21781005ce5005191406e9bf9f

tests/integration/arlocal.int.test.ts
matibat: e5c26af1d4daaa86c242439106cb93445a7df9a7

src/types/custom_metadata_types.ts
matibat: I did: 8ae049ea8e1e4c6318453e3c495a66c014a693f7

tests/integration/arlocal.int.test.ts
```diff
@@ -486,6 +486,37 @@ describe('ArLocal Integration Tests', function () {
 			});
 		});
 
+		it('we can upload a public file as a v2 transaction with custom data gql tags', async () => {
+			const fileName = 'custom_content_unique_stub';
```
fedellen, line 490: We should use unique names on each test to avoid revision collisions on ArLocal

fedellen: Also, it doesn't look like we need to preserve the file name variable for this set of tests

tests/integration/arlocal.int.test.ts
```diff
@@ -1220,6 +1251,38 @@ describe('ArLocal Integration Tests', function () {
...
 		});
 
+		it('we can upload a private file as a v2 transaction with custom data gql tags', async () => {
+			const fileName = 'custom_content_unique_stub';
+			const customMetaData: CustomMetaData = {
+				dataGqlTags: {
+					'My-Tag-1': 'My awesome value',
+					'My-Tag-2': ['hello', 'world!']
+				}
+			};
```
fedellen, line 1261: We could share these with the public test as `customDataTxGqlTags`

Or even make them unique per test instead

fedellen (approved):
Looking great! Left some non blocking comments

tests/integration/arlocal.int.test.ts
matibat: U right, I'm now not specifying the name but instead using two different files with distinct names
236ae54b6040e1d44ef3fdfbeddb94c4358b67c5

tests/integration/arlocal.int.test.ts
matibat: Unique tags: 236ae54b6040e1d44ef3fdfbeddb94c4358b67c5

fedellen (approved):

matibat:
Alpha version published as: ardrive-core-js@1.16.0-alpha-2

matibat:
Oops, I released the alpha again. It's now: ardrive-core-js@1.17.0-alpha-0

PR #208, opened 2022-07-28
PE-1972: protect all arfs tags

fedellen:


matibat (approved):
Perfect!

src/types/custom_metadata_type.test.ts
```diff
@@ -0,0 +1,11 @@
+import { expect } from 'chai';
+import { gqlTagNameArray } from '../utils/constants';
+import { isCustomMetaDataGqlTags } from './custom_metadata_types';
+
+describe('isCustomMetaDataGqlTags type guard function', () => {
+	describe('returns false if provided tag name collides with any protected ArFS GQL tag name', () => {
+		for (const tagName of gqlTagNameArray) {
```
arielmelendez, line 7: This is an example of the test code using app code in a way that could lead to false positives in our tests. What do you think about keeping a separate list here and have a test that verifies that the app's constant has the same values and array length? Then tests will break when someone adds or removes something from the app code's list.

fedellen: Ah, thats completely right

fedellen: 1c03446

matibat (approved):
LGTM

arielmelendez (approved):

arielmelendez (approved):

PR #220, opened 2023-08-10
Release ArDrive Core JS 2.0.0

fedellen:
This Release PR includes:

- Bumps ArDrive to Node 18.17.0
- Optionally send ArFS Entities to Turbo based on presence of `TurboSettings`


package.json
```diff
@@ -1,9 +1,12 @@
 {
 	"name": "ardrive-core-js",
-	"version": "1.18.0",
+	"version": "2.0.0",
 	"description": "ArDrive Core contains the essential back end application features to support the ArDrive CLI and Desktop apps, such as file management, Permaweb upload/download, wallet management and other common functions.",
 	"main": "./lib/exports.js",
 	"types": "./lib/exports.d.ts",
+	"engines": {
+		"node": ">=18.17.0"
+	},
```
fedellen, line 9: I added this as a safety net for preventing users from generating wallets with olde node versions. but it might be very strict for allowing access to the library. 

Do we want this engine restriction? @dtfiedler @karlprieb @arielmelendez

dtfiedler: i'm in favor of adding it - although we could use just the major version `>=18`

dtfiedler: it'll also protect us from devs using older node versions and hitting errors we have yet to see and having to respond to in discord/elsewhere to upgrade their node version and try again

dtfiedler (approved):

dtfiedler (approved):

PR #224, opened 2023-09-22
chore(PE-4646): replace ArFSEntityCache with @ardrive/promise-cache

dtfiedler:
…brary


src/types/ar.ts
```diff
@@ -7,7 +7,7 @@ export class AR {
 	static from(arValue: BigNumber.Value): AR {
 		const bigWinston = new BigNumber(arValue).shiftedBy(12);
 		const numDecimalPlaces = bigWinston.decimalPlaces();
-		if (numDecimalPlaces > 0) {
+		if (numDecimalPlaces && numDecimalPlaces > 0) {
```
arielmelendez, line 10: Did TS or other tooling give us a warning here?

dtfiedler: yeah - looks like the bump to bigNumber.js changed the result type of `decimalPlaces()`

src/arfs/arfs_entity_cache.ts
```diff
@@ -1,36 +0,0 @@
...
-
-	cacheKeyString(key: K): string {
-		// Note: This implementation may not sufficiently differentiate keys
-		// for certain object types depending on their toJSON implementation
-		return typeof key === 'string' ? key : JSON.stringify(key);
-	}
-
-	put(key: K, value: Promise<V>): Promise<V> {
-		this.cache.write(this.cacheKeyString(key), value);
-		return value;
```
arielmelendez, line 18: no await 😬

arielmelendez (dismissed):

dtfiedler:
Interesting that an integration test is failing - looking into it

dtfiedler:
i was experiencing some really odd behavior in the tests - i think there is some work to be done to upgrade other dependencies that are outdated (e.g. `axios`). removing those in this PR sorted out the tests, but i worry there is trouble ahead.

arielmelendez (approved):

PR #227, opened 2023-10-02
chore: update dependencies

dtfiedler:
Axios is the only major version bump in `dependencies` - the rest are devDeps or types.


package.json
```diff
@@ -10,58 +10,58 @@
 	"dependencies": {
 		"@alexsasharegan/simple-cache": "^3.3.3",
 		"@ardrive/ardrive-promise-cache": "^1.1.4",
-		"@ardrive/turbo-sdk": "^1.0.0",
-		"arbundles": "^0.6.19",
-		"arweave": "^1.11.4",
+		"@ardrive/turbo-sdk": "^1.0.1",
+		"arbundles": "^0.6.22",
+		"arweave": "^1.14.4",
```
fedellen, line 15: We could consider locking the versions of some of these most critically-used libraries

fedellen (approved):

package.json
dtfiedler: probably a good idea

PR #228, opened 2023-10-26
fix(seed phrase): use human crypto key with a set bip 39 version for generating keypair from seed phrase 

fedellen:
This PR fixes the non-deterministic and incorrect key pair being generated from seed phrases.

It seems that our issue is caused only when the package `bip39` is pulled in at version `3.1.0`. I have not solved why this happens... it seems the seed coming from `bip39` is correct every time. The package that `bip39` now uses is `@noble/hashes`. My current theory is that package is manipulating some global cryptos that `human-crypto-keys` is depending on

The PR corrects this defect by just using `human-crypto-keys` at a version that is directly locking `bip39` at `3.0.2`. This ensures that our dependencies will pull in `bip39` at the correct version for our use case no matter their install


src/wallet_dao.ts
```diff
@@ -33,13 +34,24 @@ export class WalletDAO {
...
+		const keys = await generateKeyPair({ id: 'rsa', modulusLength: 4096 }, { privateKeyFormat: 'pkcs8-pem' });
+		return new SeedPhrase(keys.mnemonic);
 	}
 
 	async generateJWKWallet(seedPhrase: SeedPhrase): Promise<JWKWallet> {
-		const jwkWallet: JWKInterface = await mnemonicKeys.getKeyFromMnemonic(seedPhrase.toString());
-		return Promise.resolve(new JWKWallet(jwkWallet));
+		const { privateKey } = await getKeyPairFromMnemonic(
+			seedPhrase.toString(),
+			{
```
arielmelendez, line 44: Nit: Might want to factor this out to a constant since it's used twice.

package.json
```diff
@@ -11,13 +11,13 @@
...
 		"arbundles": "^0.6.19",
 		"arweave": "^1.11.4",
-		"arweave-mnemonic-keys": "^0.0.9",
 		"axios": "^0.21.1",
 		"axios-retry": "^3.6.0",
 		"base64-js": "^1.5.1",
 		"bignumber.js": "^9.0.1",
 		"bn.js": "^5.2.1",
 		"futoin-hkdf": "^1.3.3",
+		"human-crypto-keys": "git+https://github.com/fedellen/js-human-crypto-keys.git#expose_lib",
```
arielmelendez, line 20: Let's move this fork to ardriveapp org.

fedellen: 5aa4132

src/wallet_dao.ts
fedellen: feefca4

arielmelendez (approved):

PR #230, opened 2023-11-10
fix(contract reader): switch to a PDS managed default contract cache

fedellen:
this PR fixes an issue we're seeing with the default contract reader when fetching the ardrive contract when determining the community tip during uploads

as we were forewarned, it seems the Verto contract cache is no longer publicly available. fortunately PDS is now hosting our own contract cache, the ArNS Microservice -- here we switch to using that cache

### DIFF
--- a/src/community/arns_microservice_contract_oracle.ts
+++ b/src/community/arns_microservice_contract_oracle.ts
@@ -4,12 +4,12 @@ import { ContractReader } from './contract_oracle';
 
 /**
  *  Oracle class responsible for retrieving and
- *  reading Smartweave Contracts from the Verto cache
+ *  reading Smartweave Contracts from the ArNS Microservice
  */
-export class VertoContractReader implements ContractReader {
+export class ArNSMicroserviceContractReader implements ContractReader {
 	/** Fetches smartweave contracts from the Verto cache */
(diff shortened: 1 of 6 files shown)


src/community/arns_microservice_contract_oracle.ts
```diff
@@ -4,12 +4,12 @@ import { ContractReader } from './contract_oracle';
 
 /**
  *  Oracle class responsible for retrieving and
- *  reading Smartweave Contracts from the Verto cache
+ *  reading Smartweave Contracts from the ArNS Microservice
  */
-export class VertoContractReader implements ContractReader {
+export class ArNSMicroserviceContractReader implements ContractReader {
```
arielmelendez, line 9: It's a generalized contract cache with some ArNS-specific APIs available. Can we just refer to this as `PDSContractCacheServiceContractReader`?

arielmelendez: I can see how the domain name might push us in this direction though...

fedellen: Sounds good

arielmelendez (approved):

arielmelendez (approved):

PR #240, opened 2024-06-03
feat(gql): utilize drive ID in conflict resolution and list folder qu…

fedellen:
This PR ensures Drive ID and owner are included in GQL read queries for indexing performance

### DIFF
--- a/src/ardrive.ts
+++ b/src/ardrive.ts
@@ -191,7 +191,11 @@ export class ArDrive extends ArDriveAnonymous {
 		}
 
 		// Assert that there are no duplicate names in the destination folder
-		const entityNamesInParentFolder = await this.arFsDao.getPublicEntityNamesInFolder(newParentFolderId, owner);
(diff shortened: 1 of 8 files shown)


src/ardrive.ts
```diff
@@ -729,6 +739,7 @@ export class ArDrive extends ArDriveAnonymous {
 		destinationFolderId
 	}: RetryPublicArFSFileByDestFolderIdParams): Promise<ArFSResult> {
 		const metaDataTx = await this.deriveMetaDataTxFromPublicFolder(destinationFolderId, dataTxId);
+		const driveId = await this.arFsDao.getDriveIdForFolderId(destinationFolderId);
```
arielmelendez, line 742: Is this a slow query these days? Also can we drill driveId down through the `retryPublicArFSFileUploadByDestFolderId` API (at the risk of making it harder to use correctly) rather than fetching it here? Note to self to double check whether that fetch result is cached...

arielmelendez (approved):

PR #242, opened 2024-07-18
PE-6440: fix(uploads): verify drive privacy

thiagocarvalhodev:
#### Summary
This PR ensures that privacy settings are validated before creating or uploading folders and files.

#### Details
If a user attempts to upload a private file to a public drive, an error will be displayed.

Example of the error message:image. 

**Other flows like rename or move files and folders are not affected.**

Fixes: https://github.com/ardriveapp/ardrive-cli/issues/352.

### DIFF
--- a/src/ardrive.ts
+++ b/src/ardrive.ts
@@ -600,6 +600,7 @@ export class ArDrive extends ArDriveAnonymous {
 			const destDriveId = await this.arFsDao.getDriveIdForFolderId(destFolderId);
 			const owner = await this.wallet.getAddress();
 
+			// Assert that the drive has the correct privacy settings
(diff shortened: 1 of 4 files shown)


src/utils/constants.ts
```diff
@@ -93,3 +93,5 @@ export const gqlTagNameRecord = {
 
 export const gqlTagNameArray = Object.values(gqlTagNameRecord);
 export type GqlTagName = typeof gqlTagNameArray[number];
+
+export const drivePrivacyTagName = 'Drive-Privacy';
```
fedellen, line 97: nit: I would share the string from the above `gqlTagNameRecord` here just to reduce the times we declare it

PR #243, opened 2024-07-22
fix(list-folders)

thiagocarvalhodev:
### Description
This PR introduces a mechanism to skip entities that are broken due to an incorrect privacy state. When an entity is determined to be broken, a log entry will be created in the terminal to notify the user of the issue.

1. Added logging functionality to output broken entity information to the 
2. Example log entry:

image

### DIFF
--- a/src/arfs/arfsdao.ts
+++ b/src/arfs/arfsdao.ts
@@ -1528,16 +1528,27 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 			const { edges } = transactions;
 			hasNextPage = transactions.pageInfo.hasNextPage;
 
-			const folders: Promise<ArFSPrivateFolder>[] = edges.map(async (edge: GQLEdgeInterface) => {
-				cursor = edge.cursor;
-				const { node } = edge;
-				const folderBuilder = ArFSPrivateFolderBuilder.fromArweaveNode(node, this.gatewayApi, driveKey);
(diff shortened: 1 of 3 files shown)


src/arfs/arfsdao.ts
```diff
@@ -1528,16 +1528,31 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
+					// Build the folder so that we don't add something invalid to the cache
+					const folder = await folderBuilder.build(node);
+					const cacheKey = { folderId: folder.entityId, owner, driveKey };
+					return this.caches.privateFolderCache.put(cacheKey, Promise.resolve(folder));
+				} catch (e) {
+					// If the folder is broken, skip it
+					if (e instanceof SyntaxError) {
+						console.error(`Error building folder. Skipping... Error: ${e}`);
+						return null;
+					}
```
fedellen, line 1545: This fix is focused on the list command. Consider bringing this down into arfs_file_builder.ts and folder_builder directly where we're running the error causing JSON.parse. That way it can fix the issue for upload file and other commands this could be effecting

fedellen: sounds good. I'm just worried here about those other commands being effected

fedellen (dismissed):
the fix for list-folder on drive's with broken states is worth going out -- approving here

fedellen (approved):

PR #247, opened 2025-05-16
PE-8050: CLI update for ArFS v0.15

kunstmusik:
* Updates gh actions to v4 to fix issues with cache service 
* Updates dependencies to dha-team/arbundles and updates arweave to latest
* Updates yarn sdks to latest

* Updates ArFS version to v0.15
* Adds DriveSignature to private Drive entities
* New getDriveSignatureInfo() function to check drives to get `{ driveSignatureType, encryptedSignatureData? }
* Updates deriveDriveKey to work with v1 and v2 drive signatures as well as downloaded encrypted signatures
* Introduces DriveKey class to bundle driveSignatureType with key data


src/utils/crypto.ts
```diff
@@ -26,19 +27,76 @@ export async function getArweaveWalletSigningKey(jwk: JWK, data: Uint8Array): Pr
...
 // Derive a key from the user's ArDrive ID, JWK and Data Encryption Password (also their login password)
+// Defaults to DriveSignatureType.v1 for backwards compatibility.
+// ArFS v0.15 introduced stored encrypted v1 signatures which can be passed in here for decryption using a
+// v2 Drive key. Private Drives may also be created using v2 signatures. The drive's signature type is stored
+// as "Signature"
 export async function deriveDriveKey(
 	dataEncryptionKey: crypto.BinaryLike,
 	driveId: string,
-	walletPrivateKey: string
+	walletPrivateKey: string,
```
arielmelendez, line 60: Good opportunity to move to using an object parameter here to prevent walletPrivateKey being passed in at the driveId position.

src/utils/crypto.ts
```diff
@@ -26,19 +27,76 @@ export async function getArweaveWalletSigningKey(jwk: JWK, data: Uint8Array): Pr
...
+	}
 ): Promise<DriveKey> {
 	const driveIdBytes: Buffer = Buffer.from(parse(driveId) as Uint8Array); // The UUID of the driveId is the SALT used for the drive key
 	const driveBuffer: Buffer = Buffer.from(utf8.encode('drive'));
 	const signingKey: Buffer = Buffer.concat([driveBuffer, driveIdBytes]);
-	const walletSignature: Uint8Array = await getArweaveWalletSigningKey(JSON.parse(walletPrivateKey), signingKey);
+
+	let walletSignature: Uint8Array;
+
+	if (signatureType === DriveSignatureType.v1) {
```
arielmelendez, line 73: Nit: using a switch here might get us more compile-time support if we ever add another sig type

src/utils/crypto.ts
arielmelendez, line 60: I know this is historical code but it seems silly to saddle this function with deserializing the JWK from a string. All but one caller of this function are serializing the JWK before calling it. Would make more sense to me for it to take a JWK and in the singular instance where it's using a JWK STRING, deserialize it THERE before calling this function.

src/utils/crypto.ts
```diff
@@ -26,19 +27,76 @@ export async function getArweaveWalletSigningKey(jwk: JWK, data: Uint8Array): Pr
...
+				encryptedSignatureData.cipherIV,
+				driveKeyV2,
+				encryptedSignatureData.encryptedData
+			);
+		} else {
+			walletSignature = await generateWalletSignatureV1(JSON.parse(walletPrivateKey), signingKey);
+		}
+	} else if (signatureType === DriveSignatureType.v2) {
+		// v2 signature uses equivalent to Wander's signDataItem()
+		walletSignature = await generateWalletSignatureV2(JSON.parse(walletPrivateKey), signingKey);
```
arielmelendez, line 92: In the absence of a tighter type-based constraint on logic here, we might want to add an assertion or error here that encryptedSignatureData has NOT been provided.

src/arfs/arfs_builders/arfs_drive_builders.ts
```diff
@@ -151,7 +165,17 @@ export class ArFSPrivateDriveBuilder extends ArFSDriveBuilder<ArFSPrivateDrive>
 		if (!driveId) {
 			throw new Error('Drive-ID tag missing!');
 		}
-		const fileBuilder = new ArFSPrivateDriveBuilder({ entityId: EID(driveId), key: driveKey, gatewayApi });
+
+		const driveSignatureTypeTagData = tags.find((tag) => tag.name === 'Signature-Type')?.value;
+		const driveSignatureType = !driveSignatureTypeTagData
```
arielmelendez, line 170: nit: prefer the positive case logic first, i.e. `driveSignatureTypeTagData ? parseDriveSignatureType(driveSignatureTypeTagData) : undefined`

tests/integration/ardrive.int.test.ts
```diff
@@ -440,7 +440,7 @@ describe('ArDrive class - integrated', () => {
 
 				assertCreateDriveExpectations({
 					result,
-					driveFee: W(2973),
+					driveFee: W(2990),
```
arielmelendez, line 443: Why did this change? Is it because there's more tag data?

src/ardrive.ts
```diff
@@ -1917,4 +1920,8 @@ export class ArDrive extends ArDriveAnonymous {
 
 		return this.arFsDao.downloadPrivateFolder(downloadFolderArgs);
 	}
+
+	async getDriveSignatureInfo({ driveId, owner }: GetDriveSignatureInfoParameters): Promise<DriveSignatureInfo> {
```
arielmelendez, line 1924: This codebase overuses type definitions for input object arguments. I think we should only use them when we're extending the interfaces for reuse elsewhere. Otherwise we're just creating indirection that slows down the reader. Excise this type unless you believe there is an important reason to keep it.

arielmelendez (approved):
Left some minor suggestions but the major pieces look good to go. Thanks!

src/utils/crypto.ts
kunstmusik: When I started on this I wanted to keep changes to a minimum but I think at this point might as well go for it. :) Will update the function signature and report back the commit.

arielmelendez (approved):

arielmelendez (dismissed):

PR #249, opened 2025-06-02
PE-8050: restore backwards compatibility

arielmelendez:


fedellen (approved):
this part looks good! 

will continue review on the above PR

PR #253, opened 2025-06-23
feat: add Ethereum signer support for Turbo integration

fedellen:
- Upgrade @ardrive/turbo-sdk from v1.0.1 to v1.27.0 (pinned)
- Add support for Ethereum private keys and custom signers in Turbo
- Enhance Turbo class to support both authenticated and unauthenticated clients
- Add ethereumPrivateKey and turboSigner parameters to ArDrive factory
- Implement private key validation with proper error handling
- Add comprehensive unit and integration tests
- Improve code clarity with better documentation and helper functions
- Ensure backwards compatibility for existing implementations

BREAKING CHANGES: None - all changes are backwards compatible


src/ardrive_factory.ts
```diff
@@ -78,34 +128,36 @@ export function arDriveFactory({
 	shouldBundle = true,
 	arFSTagSettings = new ArFSTagSettings({ appName, appVersion }),
 	turboSettings = undefined,
+	ethereumPrivateKey = undefined,
+	turboSigner = undefined,
 	uploadPlanner = new ArFSUploadPlanner({
 		shouldBundle,
 		arFSTagSettings,
-		useTurbo: !!turboSettings
+		useTurbo: !!turboSettings || !!ethereumPrivateKey || !!turboSigner
```
fedellen, line 136: the planner doesn't necessarily need these new booleans. the intent is that when no turbo settings exist, we should not use turbo

src/ardrive_factory.ts
```diff
@@ -51,6 +52,18 @@ export interface ArDriveSettings extends ArDriveSettingsAnonymous {
...
+	 * Takes precedence over turboSigner if both are provided.
+	 */
+	ethereumPrivateKey?: string;
+	/**
+	 * Custom signer for Turbo operations.
+	 * Can be any supported signer type (EthereumSigner, ArweaveSigner, etc).
+	 * Ignored if ethereumPrivateKey is provided.
+	 */
+	turboSigner?: TurboSigner;
 }
```
fedellen, line 67: these are duplicates. don't think we need this second set of params. the ones added to  turboSettings should handle this

tests/integration/turbo-ethereum.int.test.ts
```diff
@@ -0,0 +1,293 @@
...
+import { ArDrive } from '../../src/ardrive';
+import { arDriveFactory } from '../../src/ardrive_factory';
+import { Turbo } from '../../src/arfs/turbo';
+import { wrapFileOrFolder } from '../../src/arfs/arfs_file_wrapper';
+import { EthereumSigner } from '@ardrive/turbo-sdk';
+import { JWKWallet } from '../../src/jwk_wallet';
+import { readJWKFile } from '../../src/utils/common';
+import { EID, FolderID } from '../../src/types';
+
+describe('Turbo Ethereum Signer Integration Tests', function () {
```
fedellen, line 13: getting failures on this suite. I'm working on getting these tests working with the proper stubs

src/ardrive_factory.ts
```diff
@@ -65,6 +78,43 @@ const defaultTurboSettings = {
...
+			...turboConfig,
+			ethereumPrivateKey,
+			signer: turboSigner,
+			useAuthenticated: true
+		});
+	}
+
+	// Return unauthenticated Turbo
+	return new Turbo(turboConfig);
+}
```
fedellen, line 116: ah. the reason we don't use authenticated turbo here in core js is because we're not using turbo sdk to sign data items. in order to sign the data items the PR would need to effect arfsdao.ts

PR #257, opened 2025-07-02
Add Ethereum wallet support

fedellen:
this PR adds ETH Wallet Support for All ArDrive Core Write and Read interactions

specifically:

- extend `Wallet` interface with transaction and data item signing hooks and `getAllAddresses` capability
- implement `EthereumWallet` using ethers and arweave-js ECDSA Release
- update `JWKWallet` to comply with new interface
- adapt transaction preparation and transfers to use new wallet methods

Notably L1 and L2 Addresses for the ETH signers are different here. This is because of the difference the way that Arweave layer 1 nodes and arbundles layer 2 data items derive and expect ethereum public keys. ANS-104 uses uncompressed public keys, while Arweave Nodes expect compressed public keys. To circumvent this we use a getAllAddresses() helper that gathers all potential owners for the wallet, so when we read GQL we pass both owners to the query


src/arfs/arfsdao.ts
```diff
@@ -1502,6 +1505,7 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 
 		const driveTransactions = await this.gatewayApi.gqlRequest(gqlDriveQuery);
 
+		console.log('driveTransactions', driveTransactions);
```
fedellen, line 1508: TODO: rogue console logs in this whole file. might as well let husky stop console.log in this repo. 

another thought, would be nice to have a `logger` drilled around this library for leaving debug logs. maybe we can leverage TurboLogger 🤔

src/wallet_dao.test.ts
```diff
@@ -5,8 +5,8 @@ import { WalletDAO } from './wallet_dao';
 
 // This test runs too slow to be included in the CI pipeline. But it provides a snapshot of the
 // seed phrase to wallet functionality and should be run locally before a release
-describe.skip('Wallet DAO', function () {
-	this.timeout(90_000);
+describe('Wallet DAO', function () {
+	this.timeout(90_000000);
 	const walletDAO = new WalletDAO(fakeArweave);
```
fedellen, line 10: TODO: unskip -- this was for testing any getAddress breaking changes

src/ethereum_wallet.ts
```diff
@@ -0,0 +1,111 @@
...
+		);
+		return ethNormalizedL1Address;
+	}
+
+	/**
+	 * Returns an object containing the normalized Arweave owners for the wallet.
+	 * - `networkAddress`: L1 arweave-js normalized address
+	 * - `ans104Address`: L2 arbundles normalized address
+	 */
+	// TODO: change to  or similar
```
fedellen, line 67:
```suggestion
```

src/arfs/arfsdao_anonymous.ts
```diff
@@ -333,7 +333,7 @@ export class ArFSDAOAnonymous extends ArFSDAOType {
 				try {
 					const folderBuilder = ArFSPublicFolderBuilder.fromArweaveNode(node, this.gatewayApi);
 					const folder = await folderBuilder.build(node);
-					const cacheKey = { folderId: folder.entityId, owner };
+					const cacheKey = { folderId: folder.entityId, owner: Array.isArray(owner) ? owner[0] : owner };
```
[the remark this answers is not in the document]
fedellen: so maybe the fn for backwards compat can take one address, or the new object type

src/ethereum_wallet.ts
```diff
@@ -0,0 +1,110 @@
...
+
+import { PublicKey, ArweaveAddress, ADDR } from './types';
+import { bufferTob64Url, ownerToAddress, toB64Url } from './utils/wallet_utils';
+import { Wallet } from './wallet';
+import { ec as EC } from 'elliptic';
+import Arweave from 'arweave';
+import { fromJWK, SECP256k1PublicKey } from 'arweave/node/lib/crypto/keys';
+import { JWKInterface } from 'arweave/node/lib/wallet';
+
+// TODO: check for new arweave js ecdsa releases
```
fedellen, line 12: checked, this is the latest release


```suggestion
```

src/arfs/arfsdao_anonymous.ts
```diff
@@ -163,8 +169,8 @@ export class ArFSDAOAnonymous extends ArFSDAOType {
 	}
 
 	// Convenience function for known-public use cases
-	async getPublicDrive(driveId: DriveID, owner: ArweaveAddress): Promise<ArFSPublicDrive> {
-		const cacheKey = { driveId, owner };
+	async getPublicDrive(driveId: DriveID, owner: ArweaveAddress | WalletAddresses): Promise<ArFSPublicDrive> {
+		const cacheKey = { driveId, owner: owner instanceof ArweaveAddress ? owner : owner.networkAddress };
```
fedellen, line 173: nit: can use helper

PR #262, opened 2025-09-12
Optimizations for list folder

kunstmusik:


fedellen (approved):

vilenarios:
Thanks @kunstmusik — closing this as **superseded**, with the good ideas carried forward.

Two things were bundled here:

1. **Single-query folder-children consolidation** (the `getPublicFilesWithParentFolderIds(searchFolderIDs, …)` batching) — this is now handled on `master` and expanded in #275 (multi-value tag queries + page size 100→1000 + bounded fetch concurrency; [benchmark: ~3.7× fewer round trips](https://github.com/ardriveapp/ardrive-core-js/pull/275)). The 4.1.0 listing rewrite is also why this branch now conflicts with `master`.

2. **Parallel folder downloads** — a genuinely good idea we want to keep. One caveat with the approach here: `childrenFiles.map(async …)` starts *every* download immediately (each async body runs to its first `await` during `.map`), so `Promise.all` in batches of 5 only *awaits* in groups — it doesn't actually cap in-flight downloads at 5. We're re-doing this on current `master` using the new bounded-concurrency helper so it truly limits concurrency, with a test that asserts the in-flight cap.

Appreciate the contribution — the download-parallelism follow-up PR will credit this. 🙏

PR #263, opened 2025-09-26
PE-8597: browser build and tests

kunstmusik:
## Summary by CodeRabbit

* **New Features**
  * Browser-compatible web build module with signer-based wallet integration
  * Support for private drive operations in web builds
  * New web module exports for browser environments

* **Documentation**
  * Added comprehensive guides for web builds and browser wallet integration

* **Tests**
  * Implemented Playwright browser testing suite

* **Chores**
  * Version updated to 3.1.0
  * Updated build tooling and dependencies


src/ardrive.ts
```diff
@@ -121,9 +124,11 @@ import {
 } from './arfs/arfs_entity_name_validators';
 
 export class ArDrive extends ArDriveAnonymous {
+	private readonly signer?: ArweaveSigner;
+
 	constructor(
-		private readonly wallet: Wallet,
-		private readonly walletDao: WalletDAO,
+		private readonly wallet: Wallet | undefined,
+		private readonly walletDao: IWalletDAO,
```
arielmelendez: It seems we created this abstraction (IWalletDAO) but then never end up using it. Is it still necessary? I think the change could be fine as long as the change is backwards compatible for the primary interfaces that 90+% of users are using (e.g. ArDriveFactory).

kunstmusik: Looks like this was work from an iteration that went down a path that was superceded later. Working on removing this interface and web implementation now.

src/utils/common_browser.ts
```diff
@@ -0,0 +1,158 @@
+/**
+ * Browser-compatible utilities extracted from common.ts
+ * These functions have no Node.js dependencies and can be used in both Node.js and browser environments
+ */
+
+import { ByteCount, DriveSignatureType } from '../types';
+import { authTagLength } from './constants';
+
+/** Derives gateway URL from provided Arweave instance */
+export function gatewayUrlForArweave(arweave: {
```
arielmelendez: Did all of these really need to be copied over? Could we have managed to simply split up files more and/or make some of the interfaces used more cross-platform friendly?

src/ardrive.ts
kunstmusik: Removed the interface in f9b77d7127506a62c1647344b4359f9475f64811

src/utils/common_browser.ts
kunstmusik: These are split off from common.ts as those which are safe to use from both browser and node, but probably shouldn't be defined in both. I'm reviewing now to see if there's any duplicates and removing the safe ones from common.ts. Claude is going down a path of re-exporting functions from common_browser within common.ts, for backwards compatibility. I think that's a good step to remove repetition.

src/web/ardrive_factory_web.ts
```diff
@@ -0,0 +1,92 @@
...
+export function arDriveAnonymousFactory({
+	gatewayUrl = new URL('https://arweave.net/')
+}: ArDriveSettingsAnonymousWeb = {}) {
+	// Create GatewayAPI for web
+	const gatewayApi = new GatewayAPI({ gatewayUrl });
+
+	// Create ArFSDAOAnonymous with minimal dependencies
+	// We pass null for arweave since web doesn't use Node.js-specific methods
+	const dao = new ArFSDAOAnonymous(
+		null as unknown as Arweave, // arweave - not used by web methods
```
arielmelendez: Fine, though might have preferred a backwards-compatible update to the ArFSDAOAnonymous ctor where this field is "Arweavelike" and then you could pass a FakeArweave implementation here throws if you attempt to actually use it.

src/web/community_oracle_web.ts
```diff
@@ -0,0 +1,29 @@
...
+		return Promise.resolve(minArDriveCommunityWinstonTip);
+	}
+
+	/**
+	 * Returns a default ArDrive community address for browser environments
+	 */
+	async selectTokenHolder(): Promise<ArweaveAddress> {
+		// Return the ArDrive community address as default
+		// This is a placeholder - in production, this should be fetched from a contract oracle
+		return Promise.resolve(ADDR('vLRHFqCw1uHu75xqB4fCDW-QxpkpJxBtFD9g4QYUbfw'));
```
arielmelendez: Does this code actually run in web builds?

src/ardrive.ts
```diff
@@ -1377,15 +1424,42 @@ export class ArDrive extends ArDriveAnonymous {
...
+	 */
+	async signData(bytes: Uint8Array, tags: { name: string; value: string }[] = []) {
+		// Import DataItem creation utilities
+		const { ArweaveSigner, createData } = await import('@dha-team/arbundles');
+
+		// Use provided signer, or create from wallet
+		let actualSigner = this.signer;
+		if (!actualSigner && this.wallet) {
+			// eslint-disable-next-line @typescript-eslint/no-explicit-any
+			const jwkWallet = this.wallet as any;
```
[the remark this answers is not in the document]
kunstmusik: 0ff8323a2efa36f1e887450aae64161713a49bb6

src/web/turbo_web.ts
```diff
@@ -0,0 +1,73 @@
...
+			dataCaches: [],
+			fastFinalityIndexes: [],
+			winc: '0'
+		};
+		if (this.isDryRun) {
+			return defaultResponse;
+		}
+
+		// Convert DataItem buffer to ReadableStream for browser
+		const buffer = dataItem.getRaw();
```
arielmelendez: This encouraged me to check whether Node was using streaming for Turbo uploads and it's not 😢 That is to say, it's using the streaming APIs, but uses an in-memory data item. Ugh.

src/web/crypto_web.ts
```diff
@@ -0,0 +1,267 @@
...
+
+	// Build signing input: "drive" + UUID(driveId)
+	const driveBytes = new TextEncoder().encode('drive');
+	const uuidBytes = uuidParse(driveId);
+
+	// Convert to Uint8Array if needed
+	const uuidArray = uuidBytes instanceof Uint8Array ? uuidBytes : new Uint8Array(uuidBytes);
+
+	// Create sign input by concatenating drive bytes and UUID bytes
+	const signInput = new Uint8Array(driveBytes.length + uuidArray.length);
```
arielmelendez: Use driveByte.byteLength and uuidArray.byteLength

src/web/crypto_web.ts
```diff
@@ -0,0 +1,267 @@
...
+	signInput.set(driveBytes, 0);
+	signInput.set(uuidArray, driveBytes.length);
```
arielmelendez: Use driveBytes.byteLength

src/web/arfsdao_authenticated_web.ts
```diff
@@ -0,0 +1,427 @@
...
+	 * Note: Simplified implementation without caching for browser
+	 */
+	async getPrivateFile(fileId: FileID, driveKey: DriveKey, owner: ArweaveAddress): Promise<ArFSPrivateFile> {
+		return new ArFSPrivateFileBuilder(fileId, this.gatewayApi, driveKey, owner).build();
+	}
+
+	/**
+	 * Get all folders of a private drive
+	 */
+	async getAllFoldersOfPrivateDrive({
```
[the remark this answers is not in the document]
kunstmusik: In general that was the plan.

src/web/community_oracle_web.ts
kunstmusik: BTW: I think my earliest directives going through this was to make the web version as feature complete as possible with the node version, but that seems to have included thing like implementing classes which are exported by the node library that don't make sense within the web context and standard usage of the library.

src/arfs/arfs_file_wrapper.ts
```diff
@@ -149,9 +145,30 @@ export class ArFSManifestToUpload extends ArFSDataToUpload {
 		});
 
 		// TURN SORTED CHILDREN INTO MANIFEST
-		const pathMap: ManifestPathMap = {};
+		// Deduplicate files by dataTxId - keep only unique data transactions
```
arielmelendez: Sweet

src/web/turbo_web.ts
```diff
@@ -0,0 +1,73 @@
...
+		const stream = new ReadableStream({
+			start(controller) {
+				controller.enqueue(buffer);
+				controller.close();
+			}
+		});
+
+		return this.turbo.uploadSignedDataItem({
+			dataItemStreamFactory: () => stream,
+			dataItemSizeFactory: () => buffer.length
```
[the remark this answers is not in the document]
kunstmusik: Checked with Claude and it actually recommends to use length with Uint8Array, vs. byteLength. (Determined that value from arbundles is a Uint8Array).

src/web/crypto_web.ts
kunstmusik: Claued recommended to use driveBytes.length; says would have to use driveBytes.buffer.byteLength, but that since driveBytes is a view, underlying array might actually be larger in size than the view's length.

Analysis mentions:

```
TextEncoder().encode() returns a Uint8Array that might be a view into a larger buffer (implementation-dependent). Using .buffer.byteLength could give you the wrong size.

Summary
.length - Number of elements in the TypedArray view (what you actually want)
.buffer.byteLength - Size of the entire underlying buffer (may be larger than the view)
Always use .length for Uint8Array unless you specifically need to know the underlying buffer size for some reason. In your concatenation case, you definitely want .length.
```

arielmelendez: I think it's ok here as there's nothing special about the encoding of the data going into the signInput. The word "drive" and a uuid are basic ascii text so length and byteLength should be the same.

arielmelendez (dismissed):

arielmelendez (approved):

PR #271, opened 2026-07-04
fix(arfs): throw a clear error when a drive GQL query returns zero edges

vilenarios:
## Problem
`ArDrive.getPrivateDrive` (via `getDriveSignatureInfo`) and `isPublicDrive` read `edges[0].node` from the GraphQL response with **no empty-edges guard**. When a drive query returns zero edges — e.g. a wrong/missing owner, or a gateway whose index doesn't (yet) hold that drive — they throw an opaque `TypeError: Cannot read properties of undefined (reading 'node')` instead of a clear "drive not found".

Found via real on-chain testing (ArDrive Desktop): querying an incomplete-index gateway for a drive absent from that index crashed with the cryptic `TypeError` on every drive, rather than a usable error.

## Fix
Guard the empty-edges case in `getDriveSignatureInfo` and `isPublicDrive` — throw a clear, actionable error:
> `Drive with Drive-Id "<id>" not found (check the drive id and that the owner address is correct)`

The three drive builders (`ArFSPublicDriveBuilder` / `ArFSPrivateDriveBuilder` / `SafeArFSDriveBuilder`) already guard empty edges via the shared `parseFromArweaveNode`; these two standalone GQL calls were the remaining unguarded `edges[0]` sites (audited all 8 `edges[0]` usages — no others needed a guard).

## Tests
Adds 2 tests in `arfsdao.test.ts` asserting both functions throw the clear "not found" error (not `TypeError`) when the GQL query returns `edges: []`. Build + the new tests pass on top of current `master`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved drive lookup handling when no matching results are returned from the network.
  * Added clearer error messages for missing drives instead of failing on empty results.
  * Strengthened checks for both public and signature-based drive lookups.
* **Tests**
  * Added coverage for empty search results to verify the new error behavior.

### DIFF
--- a/src/arfs/arfsdao.test.ts
+++ b/src/arfs/arfsdao.test.ts
@@ -1030,6 +1030,73 @@ describe('The ArFSDAO class', () => {
 	});
 });
 
+// These GQL empty-edges guard tests live in their own describe with a minimal
+// fixture that does NOT derive a stub drive key, so they exercise the drive
+// lookup guards directly without depending on private-key crypto helpers.
+describe('The ArFSDAO class GQL empty-edges guards', () => {
+	const emptyEdgesWallet = readJWKFile('./test_wallet.json');
+	const emptyEdgesArweave = Arweave.init({
+		host: 'localhost',
+		port: 443,
+		protocol: 'https',
+		timeout: 600000
+	});
+	const emptyEdgesTagSettings = new ArFSTagSettings({ appName: 'ArFSDAO-Test', appVersion: '1.0' });
+
+	let emptyEdgesGatewayApi: GatewayAPI;
+	let emptyEdgesDao: ArFSDAO;
+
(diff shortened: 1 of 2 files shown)


vilenarios:
Thanks @coderabbitai — applied the helper extraction: added `assertDriveEdgesFound(driveId, edges)` and both `getDriveSignatureInfo` and `isPublicDrive` now call it, so the not-found message can't drift. Build + both empty-edges guard tests pass on master.

On the second nitpick (aligning phrasing with `getOwnerAndAssertDrive`'s existing `Could not find a transaction…` message): leaving that as-is to keep this PR focused, since it's a pre-existing guard with a different message and, as noted, not required here — happy to unify all drive-lookup guards in a follow-up if the team wants.

PR #275, opened 2026-07-09
perf(gql): page size 100→1000 + bounded fetch concurrency (~10× fewer GraphQL requests)

vilenarios:
## What

Raise the GraphQL page size **100 → 1000** across all listing/snapshot/incremental paths, and bound per-batch entity-fetch concurrency so the larger pages don't multiply parallel fetches.

## Why (measured)

turbo-gateway / ar.io gateways serve up to **1000 items/page** (they silently cap at 1000; `first:1001` → 1000, no error), but core-js requests 100. Live measurement on a real owner-scoped query (982 file+folder entities across 15 drives):

| page size | requests | time | entities |
|---|---|---|---|
| 100 | 10 | 3,628 ms | 982 |
| **1000** | **1** | **550 ms** | 982 |

**~10× fewer GraphQL requests, ~6.6× faster, identical result set** — which also means ~10× less rate-limit exposure.

## Changes

- **Page size (CORE-7):** new `GQL_PAGE_SIZE = 1000` constant routed through `pageLimit` (buildQuery default), both incremental-DAO `batchSize` defaults, and `SNAPSHOT_PAGE_LIMIT`. **Correctness-safe:** all pagination is `hasNextPage`-driven (never `edges.length === pageSize`), so against a gateway that returns fewer than requested it keeps following the cursor — **no entity is ever dropped** (covered by a new capped-gateway test). `batchSize` remains an additive per-call override; public API unchanged.
- **Bounded concurrency (CORE-9):** raising the page 10× would 10× the peak concurrent per-entity metadata fetches per batch. New `src/utils/concurrency.ts` (shared-cursor worker pool, input-order-preserving) caps per-batch fetches at `MAX_CONCURRENT_ENTITY_FETCHES = 30` (~3× below the old page-100 worst case) at all 16 fan-out sites. **Result set is identical** — only parallelism is bounded (index-aligned, same `Promise.all`/`allSettled` semantics, cursor/early-stop/stats untouched); a peak-in-flight test proves ≤ 30 with all entities processed.
- **Query consolidation (CORE-8):** the full-listing paths already fetch files+folders in one multi-value query. The *incremental* file/folder queries are intentionally kept separate — each carries an independent per-type `stopAfterKnownCount` early-stop, and they already run in parallel (`Promise.all`), so a merge would be a ~zero-latency win while risking that invariant.

## Follow-up (not in this PR)

**CORE-10:** the per-entity metadata fetch silently drops a transiently-failed entity (`console.warn` + `failedEntities++`, no retry) — pre-existing; the concurrency cap here restores the prior peak but a bounded retry is worth adding separately.

## Testing

- `yarn typecheck` ✅ · `yarn build` ✅ · eslint (changed files) ✅ · `npx mocha` unit suite **876 passing** (incl. new no-drop + peak-in-flight tests).
- ⚠️ Authored in a linux/arm64 sandbox — the **ArLocal integration suite** (docker) and the **drive-id-gated incremental integration suite** could not run locally. **Please let CI (`yarn ci` on amd64/docker) run them before merge.**

Adversarially reviewed for the no-dropped-entities invariant (capped-gateway proven), worker-pool correctness (run-once + input-order), and no regression to the snapshot/incremental paths.

## Summary by CodeRabbit

* **New Features**
  * Improved pagination performance for listing and sync operations, with larger page requests that reduce round trips.
  * Added bounded parallel processing to help large folders and drives load more smoothly.

* **Bug Fixes**
  * Fixed cases where pagination could stop early and miss items.
  * Preserved existing behavior for malformed or invalid entities while preventing excessive concurrent requests.

* **Documentation**
  * Updated sync and query guidance to reflect the new default page size and concurrency behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,39 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **Bounded per-batch entity-fetch concurrency (CORE-9)**: turning a page of GraphQL
+  edges into built entities does a per-entity metadata GET, and every fan-out site did
+  so for the whole page at once (`Promise.all`/`Promise.allSettled(edges.map(...))`).
+  After CORE-7 raised the page size 100 → 1000 that meant up to ~1000 concurrent
+  metadata fetches per host per batch — a ~10x jump in peak parallelism that spikes
+  memory and open connections and risks gateway rate-limits. Concurrency is now
+  DECOUPLED from page size: a new bounded-concurrency helper (`src/utils/concurrency.ts`,
+  `mapWithConcurrency` / `mapSettledWithConcurrency`) processes each batch in
+  concurrency-limited waves capped at `MAX_CONCURRENT_ENTITY_FETCHES = 30`
+  (`src/utils/constants.ts`). So a 1000-entity page still costs ~10x fewer GraphQL
+  round-trips (CORE-7) while at most 30 entity fetches are ever in flight at once
+  (~3x below the pre-CORE-7 worst case of 100). Applied at every per-entity-fetch
+  fan-out: both incremental-sync DAOs' `processFolder/FileBatch`, the full-listing
+  walks (`getAllDrivesForAddress`, `getPublic/PrivateFilesWithParentFolderIds`,
+  `getAllFoldersOfPublic/PrivateDrive`, `getEntitiesInFolder`), the snapshot-tail
+  builders (`buildPublic/PrivateFolders/FilesFromNodes` — snapshot nodes build from
+  the seeded metadata cache with no fetch, only the live tail fetches), and the web
+  authenticated DAO's private folder/file listings. The change is internal and
+  additive: the helpers preserve input order, run each item exactly once, and keep the
+  exact `Promise.all` (reject-on-any) vs `Promise.allSettled` + `failedEntities`
+  semantics of the site they replace, so the entity SET returned is identical — only
+  parallelism is bounded. Pagination, the CORE-7 `hasNextPage` safety and the #43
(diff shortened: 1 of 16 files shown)


vilenarios:
## Benchmark: `first:1000` is ~3.7x faster than `first:100`, with identical results

Measured against the production ar.io gateway `turbo-gateway.com`, listing all ArFS entities in a real **849-entity** public drive (owner-scoped `transactions` query, cursor + `hasNextPage` pagination — the same loop core-js runs). 6 interleaved rounds per page size, median reported.

| page size | round trips | wall-clock (median) | per-request latency (min/med/max) | entities |
|-----------|-------------|---------------------|-----------------------------------|----------|
| `first:100`  | **9** | **2413 ms** (2208–2691) | 148 / 272 / 522 ms | 849 |
| `first:1000` | **1** | **647 ms** (620–702)    | 620 / 646 / 702 ms | 849 |

- **9x fewer round trips**, **3.73x lower wall-clock** (median); the two ranges don't overlap, so it isn't jitter.
- **Honest tradeoff:** a 1000-item request is ~2.4x slower *per request* (more data), but you make **1** instead of **9**, so net wall-clock drops sharply.
- **Correctness:** the set of returned tx ids is **identical** between the two page sizes (849 entities, exact sorted-set match; stable across all 6 rounds; no duplicates). A perf win that changed results would not be a win — this doesn't.
- **Ceiling check:** the gateway hard-caps a page at 1000 (`first:1001/2000/5000` all return exactly 1000), so 1000 is the correct maximum.
- **Safe by construction:** core-js pagination is `hasNextPage`-driven, not `edges.length`-driven, so a larger `first:` can never drop entities.

_Independently reproduced_ on a second run: 9 → 1 round trips, identical 849-entity set (849 unique both ways). Full method + raw per-run numbers available on request.

PR #276, opened 2026-07-09
fix(list): one malformed entity no longer aborts a whole drive listing [CORE-5]

vilenarios:
## Bug

A single on-chain entity carrying a malformed `Unix-Time` tag aborts the **entire** drive listing.

While reconstructing a file/folder entity from GQL during `listPublicFolder` / `listPrivateFolder`, `arfs_builders.ts` runs `new UnixTime(+value)`. For a negative / non-integer / non-finite value the `UnixTime` constructor throws a plain `Error('Unix time must be a positive integer!')`. None of the enumeration loops' skip clauses (which catch `SyntaxError` / `InvalidFileStateException`) catch a plain `Error`, so it propagates and one bad entity kills the whole drive reconstruction. This is **gateway-independent**.

**Live evidence (2026-07-05, UAT-SYNC-READ-PERMA):** the owner's real public drive `a173761d` hard-errored mid-`listPublicFolder` with exactly this message, blocking the drive from loading at all.

## Fix

Make the LIST/enumeration path resilient to one malformed entity while keeping strict `UnixTime` validation on WRITE paths:

- Add a catchable `InvalidUnixTimeException` (`src/types/exceptions.ts`).
- `ArFSMetadataEntityBuilder.parseFromArweaveNode` now wraps `new UnixTime(+value)` and rethrows a malformed `Unix-Time` as `InvalidUnixTimeException`. WRITE paths don't go through this GQL parser (they build `UnixTime` from `Date.now()` via prototypes), so their strict validation is untouched.
- All four listing enumeration loops skip `InvalidUnixTimeException` and continue, matching the existing `SyntaxError` / `InvalidFileStateException` skip convention:
  - `getPublicFilesWithParentFolderIds`, `getAllFoldersOfPublicDrive` (`arfsdao_anonymous.ts`)
  - `getPrivateFilesWithParentFolderIds`, `getAllFoldersOfPrivateDrive` (`arfsdao.ts`)

**Resilience approach:** SKIP the one bad entity (log + continue), not coerce to a fabricated timestamp — consistent with core-js's existing broken-entity handling and honest (no invented dates).

## Test

`src/arfs/arfsdao_anonymous.test.ts` — new case: a file list containing one entity with `Unix-Time: '-1'` now returns the other two valid files instead of throwing. Modeled on the adjacent `skips invalid files and continues processing` test.

## Verify

- `yarn build`: green.
- `yarn test` (full suite): **676 passing**, incl. the new test and the pre-existing `skips invalid files` test. The lone failure is the pre-existing `ArLocal Integration Tests` `before all` hook (`fetch failed` — requires a live ArLocal server / network; environmental, unrelated to this change).

Base branch `master` (same as PR #275). **Owner-gated for merge — do not merge without owner review.**

https://claude.ai/code/session_01RW9Vws7Mo5ETnP8ZHuGaFB

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of malformed Unix time values so a single bad record no longer breaks folder or file listings.
  * Public and private drive listings now skip invalid entries and continue returning the remaining valid results.
  * Added regression coverage to ensure listings remain stable when one item contains an invalid timestamp.

### DIFF
--- a/src/arfs/arfs_builders/arfs_builders.ts
+++ b/src/arfs/arfs_builders/arfs_builders.ts
@@ -23,6 +23,7 @@ import {
 	DriveKey
 } from '../../types';
 import { GatewayAPI } from '../../utils/gateway_api';
+import { InvalidUnixTimeException } from '../../types/exceptions';
 
 export interface ArFSMetadataEntityBuilderParams {
 	entityId: AnyEntityID;
@@ -120,7 +121,17 @@ export abstract class ArFSMetadataEntityBuilder<T extends ArFSEntity> {
 					this.entityType = value as EntityType;
 					break;
 				case 'Unix-Time':
-					this.unixTime = new UnixTime(+value);
+					// A single on-chain entity carrying a malformed Unix-Time (negative /
+					// non-integer / non-finite) must not abort a whole drive listing. The
+					// UnixTime constructor throws a plain Error here; rethrow it as a
+					// catchable InvalidUnixTimeException so LIST/enumeration loops can skip
+					// just this entity. WRITE paths don't go through this parser, so their
+					// strict UnixTime validation is unaffected.
+					try {
+						this.unixTime = new UnixTime(+value);
+					} catch {
+						throw new InvalidUnixTimeException(value);
+					}
 					break;
 				case 'Boost':
 					this.boost = new FeeMultiple(+value);
--- a/src/arfs/arfsdao.ts
+++ b/src/arfs/arfsdao.ts
@@ -190,7 +190,7 @@ import {
 } from '../exports';
 import { Turbo } from './turbo';
 import { ArweaveSigner } from '@dha-team/arbundles';
-import { InvalidFileStateException } from '../types/exceptions';
+import { InvalidFileStateException, InvalidUnixTimeException } from '../types/exceptions';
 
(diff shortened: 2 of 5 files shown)


vilenarios:
Closing as **already fixed on `master`**.

While rebasing this branch onto `master` (it was cut from `dev`, 91 commits behind), I found that CORE-5 is already resolved and released:

- `bd1575b` fix(arfs): tolerate invalid entity unixTime instead of aborting drive listing [CORE-5]
- `c39eef7` merge(core-js): tolerate invalid entity unixTime [CORE-5]
- `9b7f81f` chore(release): 4.1.0 — snapshot consumption, incremental sync, unixTime + private-listing robustness

Master's fix wraps `new UnixTime(+value)` at the shared entity-parse boundary (`parseEntityUnixTime` in `arfs_builders.ts`) and **clamps** an invalid timestamp to epoch (0) so the entity still lists, with a dedicated test in `arfs_folder_builders.test.ts`. That clamp approach is preferable to this PR's skip approach for the folder case: skipping a folder with a bad Unix-Time would orphan its children in the hierarchy, whereas clamping keeps the entity (and its subtree) visible.

This branch was an independent re-implementation off the older `dev` base and adds nothing over what's already on `master`. Closing; no action needed.

PR #277, opened 2026-07-14
ci: add npm publish workflow (fixes the dist/web publish gotcha)

vilenarios:
## What
Adds `.github/workflows/publish.yml` — core-js currently has **no publish automation**; releases are a manual `npm publish`.

## Why
`npm publish` runs the `prepare` hook (`tsc` → `lib/`) but **not** `yarn build:web` (→ `dist/web/`), which the `files` array ships. A naive publish therefore ships a **stale/missing browser bundle**. This workflow always builds both and **verifies the tarball contains `lib/` + `dist/web/` before publishing**, so a broken package can't go out.

## How it works
- **Trigger:** push a `vX.Y.Z` tag (recommended), or run manually via Actions (dry-run default).
- **Steps:** install → `yarn build` → `yarn build:web` → tag/version match check → `npm pack` content verification (lib/ + dist/web/ present) → `npm publish --provenance`.
- **One-time owner setup:** add an `NPM_TOKEN` repo secret (npm automation token with publish rights).

## Notes
- Node pinned via `.nvmrc` (18.17.0), consistent with the existing test workflow.
- Owner-gated: needs the `NPM_TOKEN` secret + your review before it can publish. Does not change any existing workflow.

## Summary by CodeRabbit

* **Chores**
  * Added an automated npm publishing workflow for version releases.
  * Validates that the release tag matches the package version and that required build artifacts are present before publishing.
  * Supports manual “dry run” mode to verify release readiness without publishing.
  * Enables npm provenance for published packages.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,83 @@
+name: Publish to npm
+
+# Publishes ardrive-core-js to npm. Two ways to trigger:
+#   1. Push a version tag `vX.Y.Z` (recommended: cut from master after merge).
+#   2. Manually via the Actions tab (workflow_dispatch) with a dry_run toggle.
+#
+# Prerequisite (one-time, owner): add an `NPM_TOKEN` repo secret — an npm
+# automation token with publish rights for `ardrive-core-js`.
+#
+# Why this exists: `npm publish` runs the `prepare` hook (tsc -> lib/) but does
+# NOT build the browser bundle `dist/web/`, which the package's `files` array
+# ships. Publishing without `yarn build:web` produces a package with a stale or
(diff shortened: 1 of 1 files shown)


vilenarios:
Thanks — switched to `v*.*.*`. For the record: `v[0-9]+.[0-9]+.[0-9]+` is actually *valid* in GitHub Actions filter-pattern syntax (it's not standard glob — `+` = one-or-more of the preceding char, `[]` = char class, `.` = literal; GitHub's own cheat sheet uses `v[12].[0-9]+.[0-9]+`), so it would have matched `v4.2.0`. But since it reads like a regex and can be misread, `v*.*.*` is unambiguous — and the "verify tag matches package.json" step is the real guard on which version publishes. Fixed in the latest commit.

PR #278, opened 2026-07-14
fix(gateway): bound 429 rate-limit retries + request timeout (fixes CLI upload hang)

vilenarios:
## Evidence / repro

A Discord user hit a **"gateway timeout" / indefinite hang while uploading via the CLI**. Live-wire forensics on `GatewayAPI` (`src/utils/gateway_api.ts`) root-caused it to the **429 rate-limit retry loop**, plus two smaller robustness gaps. This PR now carries **two real fixes + one defense-in-depth**:

### (B) THE hang — an unbounded 429 rate-limit loop *(the actual bug)*

arweave.net now **rate-limits CLI traffic and returns HTTP 429 on every request**. The retry loop's 429 branch did:

```ts
if (this.lastRespStatus === rateLimitStatus) {
    await this.rateLimitThrottle(); // 60s sleep
    continue;                        // ← no counter incremented → loops FOREVER
}
```

So a persistently-429 gateway looped **forever at 60s per iteration**. The axios request `timeout` (cause C below) does **not** help this: each 429 *response* is fast — it's the 60s **SLEEP between them** that accumulates, so it needs its own bound.

### (C) Slow data-tx fetches — no request timeout

The axios instance was `axios.create({ validateStatus: undefined })` with **no `timeout`**, so a slow/hung connection (e.g. a large data-tx GET) could wait indefinitely.

### (defense-in-depth) Permanent row-limit errors were retried 8×

ar.io gateways enforce a ClickHouse **~10M-row scan limit**; a heavy/under-scoped GraphQL query returns a permanent `TOO_MANY_ROWS` (code `158`) error that was retried up to `maxRetriesPerRequest = 8`. core-js's own queries don't trip this, so it's kept as **harmless defense-in-depth**, not the reported bug.

## Fix (minimal, surgical, backward-compatible)

**1. Bounded 429 rate-limit RETRY COUNT — the fix for the hang.**
- New `DEFAULT_MAX_RATE_LIMIT_RETRIES = 5` constant + optional `maxRateLimitRetries` constructor field (mirrors `maxRetriesPerRequest`).
- The 429 branch now tracks a **separate** `rateLimitRetries` counter; once it exceeds the budget it throws a **clear, actionable** error:
  > `Gateway is rate limiting (HTTP 429) and did not recover after N retries; try a different --gateway or wait and retry.`
- The two counters stay **independent**: a 429 does **not** consume the error-retry budget and an error does **not** consume a rate-limit retry (preserves the existing "429 doesn't consume error retries" intent — just makes it *finite*).
- **Transient-throttle tolerance preserved:** a 429 that clears within the budget still **succeeds** (default `5 × 60s ≈ 5 min` — our desktop app and on-chain test harness deliberately wait out arweave.net 429s, so this must not regress).
- The throttle duration is now injectable (`rateLimitThrottleMS`, default 60s) — a small, justified testability tweak so tests and the e2e don't sleep for real minutes.

**2. Bounded request timeout — for slow data-tx fetches.**
- `DEFAULT_GATEWAY_REQUEST_TIMEOUT_MS = 60_000` + a `requestTimeoutMs` option, applied **only** to the axios instance this class creates (`axios.create({ validateStatus: undefined, timeout: requestTimeoutMs })`). A caller-supplied `axiosInstance` is used **unmodified**. A timeout is a network error, so it flows into the existing retry/backoff path (a transient slow response retries) while a truly dead connection can no longer hang forever.

**3. Row-limit error is FATAL / non-retryable (defense-in-depth).**
- `gqlRequest` now detects the ClickHouse signals — `Limit for rows or bytes to read exceeded` **OR** `TOO_MANY_ROWS` **OR** `extensions.code === '158'` — and **fails fast** with an actionable message instead of burning 8 retries.

Everything else is unchanged: statement-timeout handling, exponential backoff, `validStatusCodes`, fatal chunk-upload errors. Only additions to the public surface are the optional `maxRateLimitRetries`, `rateLimitThrottleMS`, `requestTimeoutMs` constructor fields and their constants.

> **Out of scope (tracked separately):** gateway fail-over and changing the default gateway. This PR makes a persistent 429 fail *fast and clearly* so the operator can switch gateways; it does not switch automatically.

## Tests (`src/utils/gateway_api.test.ts`, mocha/chai/sinon, fully stubbed — no live gateway)

**Rate-limit bounding (new, sinon FAKE TIMERS — no real sleeping):**
- Persistent 429 → throws the clear rate-limit error after **exactly** `maxRateLimitRetries` throttles and does **not** loop forever.
- `429, 429, 200` → **succeeds** (transient-throttle tolerance preserved, no regression).
- Rate-limit counter and error-retry counter are **independent** (a mix of 429s and 5xx errors bounds each on its own budget).
- `maxRateLimitRetries` is **configurable** via the constructor.

**Carried from the earlier commit:** `TOO_MANY_ROWS`/code-`158` fails fast without exhausting retries; default & caller-provided request timeout; externally-supplied `axiosInstance` used unmodified; transient network/timeout error retries then succeeds; statement-timeout path unchanged.

**Full end-to-end (real local HTTP 429 server, real axios, real timers, real error propagation — NOT mocked internals):**

```
=== Scenario 1: gateway returns 429 on EVERY request ===
  requests made:  4   (3 throttled + 1 that throws)
  elapsed:        806 ms   (≈ 3 × 200ms throttle; NOT infinite)
  thrown message: GQL Error: Gateway is rate limiting (HTTP 429) and did not
                  recover after 3 retries; try a different --gateway or wait and retry.
  PASS: infinite 60s-pause hang converted to a bounded, clean failure.

=== Scenario 2 (control): 429, 429, then 200 ===
  requests made:  3   (2 throttled + 1 success)
  elapsed:        444 ms
  succeeded:      true
  PASS: transient 429 waited out and the REAL path SUCCEEDED (no regression).
```

## Verification

- `yarn build` (tsc, `tsconfig.prod.json`): **green**.
- Full suite (parallel, matches `yarn test`): **874 passing, 15 pending, 1 failing** — the single failure is the `ArLocal Integration Tests` `before all` hook (`fetch failed`, requires a live local arlocal node), pre-existing and unrelated to this change.
- New GatewayAPI rate-limit suite + e2e: **all passing**.

## Impact

`GatewayAPI` is shared by **CLI, desktop, and web**, so all three benefit. Owner-gated for merge — please do not merge without owner review.

https://claude.ai/code/session_01NhthUpcW83csH1FWtaELZD

### DIFF
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -63,6 +63,16 @@ export const defaultMaxConcurrentChunks = 32;
 export const INITIAL_ERROR_DELAY = 500; // 500ms
 
 /**
+ * Default per-request timeout (in ms) applied to the GatewayAPI's default axios
+ * instance. A hung or unresponsive gateway connection fails cleanly after this
+ * window (the resulting network error flows into the normal retry/backoff path)
+ * instead of waiting forever, which previously presented to users as an
+ * indefinite upload "hang". Callers can override via the `requestTimeoutMs`
+ * GatewayAPI constructor option, or by supplying their own `axiosInstance`.
+ */
+export const DEFAULT_GATEWAY_REQUEST_TIMEOUT_MS = 60_000; // 60 seconds
+
+/**
  *  These are errors from the `/chunk` endpoint on an Arweave
  *  node that we should never try to continue on
  */
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -73,6 +73,30 @@ export const INITIAL_ERROR_DELAY = 500; // 500ms
 export const DEFAULT_GATEWAY_REQUEST_TIMEOUT_MS = 60_000; // 60 seconds
 
 /**
+ * Default pause (in ms) applied after a gateway returns an HTTP 429 (rate limit)
+ * before the next request is attempted. Overridable via the GatewayAPI
+ * `rateLimitThrottleMS` constructor option (primarily to keep tests fast).
+ */
+export const DEFAULT_RATE_LIMIT_THROTTLE_MS = 60_000; // 60 seconds
+
+/**
+ * Default maximum number of times a persistent HTTP 429 (rate limit) response is
+ * waited out before the request gives up with a clear, actionable error.
+ *
+ * This bounds the rate-limit throttle loop: a gateway that returns 429 on EVERY
+ * request (e.g. arweave.net now rate-limits CLI traffic) would otherwise pause
+ * for {@link DEFAULT_RATE_LIMIT_THROTTLE_MS} and `continue` forever — the actual
+ * "CLI hangs / upload times out" bug. With this budget the throttle is finite
+ * (default 5 waits ≈ 5 minutes of tolerance) yet a transient throttle that
+ * clears within the budget still succeeds.
+ *
+ * NOTE: this budget is SEPARATE from `maxRetriesPerRequest` — a 429 does not
+ * consume an error retry, and a non-429 error does not consume a rate-limit
+ * retry. The two counters are intentionally independent.
+ */
+export const DEFAULT_MAX_RATE_LIMIT_RETRIES = 5;
+
+/**
  *  These are errors from the `/chunk` endpoint on an Arweave
  *  node that we should never try to continue on
  */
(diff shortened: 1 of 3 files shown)


vilenarios:
@coderabbitai review

PR #282, opened 2026-08-04
fix(crypto): decrypt AES-256-CTR private files (large streamed files) + stop silent skip

vilenarios:
## What & why

ArDrive Web encrypts **large private files with AES-256-CTR** (a streamable cipher) while small files use AES-256-GCM. The chosen cipher is recorded on-chain in the data transaction's `Cipher` tag (`AES256-GCM` / `AES256-CTR`), with the nonce in `Cipher-IV`.

core-js was **hard-wired to GCM** and never read the `Cipher` tag, so it could not decrypt CTR data. Worse, the failure was **silent**: `fileDecrypt`/`driveDecrypt` returned a `Buffer.from('Error')` / `'ERROR'` sentinel, which the private-enumeration paths then treated as malformed JSON and **skipped** — so any large private file written by ArDrive Web simply *vanished* from the drive listing, with no error surfaced.

This PR reads the on-chain `Cipher` tag and adds a correct AES-256-CTR decrypt path, and it replaces the silent sentinel with a thrown, entity-identifying error that enumeration handles per-entity (one undecryptable file no longer aborts the listing).

## Changes

- **`utils/crypto.ts`** — `fileDecrypt`/`driveDecrypt` take an optional `cipher` (default `AES256-GCM`) and branch. New `aesCtrInitialCounterFromIv()` builds the 16-byte CTR counter as `nonce ‖ 0x00000000` from the 12-byte `Cipher-IV`. CTR uses **no** auth tag (matching ArDrive Web's `Mac.empty`). Shared `decryptEntityBuffer()` now **throws `EntityDecryptionError`** (new, in `types/exceptions.ts`) instead of returning the `'Error'` sentinel.
- **`utils/stream_decrypt.ts`** — `StreamDecrypt(cipherIV, fileKey, authTag: Buffer|null, cipher = 'AES256-GCM')`; branches to a CTR `Decipher` with `authTag = null`.
- **`arfs/arfsdao.ts`** — `getCipherIVOfPrivateTransactionIDs` now also reads the **data-tx** `Cipher` tag (critical: a large file has GCM *metadata* but CTR *data*). Folder-bulk and single-file (`ardrive.ts` `downloadPrivateFile`) download paths skip the auth-tag fetch for CTR and pass the cipher through. The four private-enumeration catch blocks handle `EntityDecryptionError` — log an honest `[decrypt-failed] … (cipher X)` and exclude just that entity, so the drive listing survives.
- Builders (`arfs_file_builders.ts` / `arfs_folder_builders.ts` / `arfs_drive_builders.ts`) pass each entity's own `this.cipher` into the buffer decrypt.

## Interop is independently verified

The AES-256-CTR counter construction was confirmed against ArDrive Web's primary source (`ardrive_crypto/lib/src/stream_aes.dart`): **12-byte nonce**, counter block = `nonce ‖ (offset as 4-byte BE)`, **low 32-bit big-endian** counter — matching Node's full-128-bit-block big-endian increment below the 2^32-block (64 GiB) carry boundary (the enforced upload cap is 2 GiB). A from-scratch AES-ECB oracle (not core-js, not Node) reproduces the golden ciphertext **byte-for-byte** and confirms multi-chunk counter continuity across a 4-chunk / 769 KiB input. The committed golden test asserts decrypt of that externally-verified ciphertext literal, so it fails if the counter/CTR logic regresses.

## Tests

- `utils/crypto_aes_ctr.test.ts` — golden CTR buffer decrypt (load-bearing, hardcoded ciphertext), counter-padding, CTR routing / no-tag-strip, `driveDecrypt` CTR, GCM round-trips unchanged, `EntityDecryptionError` on GCM wrong-key + CTR decipher failure, sentinel-gone.
- `utils/stream_decrypt.test.ts` — golden streaming CTR + multi-chunk counter continuity.
- `arfs/arfsdao.test.ts` — an `EntityDecryptionError` is surfaced per-entity and the drive listing survives.

## Security note (by ArDrive Web's design, not introduced here)

AES-256-CTR is **unauthenticated** (`Mac.empty` in ArDrive Web). A wrong-key/corrupt CTR *data* file cannot be detected at decrypt time and yields garbage bytes; this is inherent to the ArDrive Web streaming format and is matched here for interop. GCM metadata remains authenticated and throws on tamper.

## Remaining confirmation

Unit vectors + the independent oracle prove the math. The one thing not exercisable without network/funds/Dart is a **live** end-to-end download of a real ArDrive-Web-written large private file — recommended as a final belt-and-suspenders check before relying on this in production.

## Summary by CodeRabbit

* **New Features**
  * Added support for decrypting private files, folders, and drives with AES-256-CTR and AES-256-GCM.
  * Improved compatibility with legacy encrypted data by defaulting missing cipher information to GCM.
* **Bug Fixes**
  * Private downloads no longer require authentication tags for CTR encryption.
  * Decryption failures now provide clear typed errors instead of invalid placeholder data.
  * Private listings continue loading when individual entities cannot be decrypted.

### DIFF
--- a/src/ardrive.ts
+++ b/src/ardrive.ts
@@ -1533,7 +1533,9 @@ export class ArDrive extends ArDriveAnonymous {
 		const fileKey = await deriveFileKey(`${fileId}`, driveKey);
 		// Read the DATA transaction's own Cipher/Cipher-IV. ardrive-web encrypts streamed (large)
 		// file data with AES256-CTR (no auth tag) even when the file's metadata is AES256-GCM, so
-		// branch on the data tx's cipher: CTR has no auth tag to fetch (absent Cipher => GCM).
+		// branch on the data tx's cipher: CTR has no auth tag to fetch (absent Cipher => GCM). The
+		// data stream itself still covers the full CTR ciphertext — the +16/-16 range cancellation is
+		// documented in getPrivateDataStream.
 		const [dataTxCipherResult] = await this.arFsDao.getCipherIVOfPrivateTransactionIDs([privateFile.dataTxId]);
 		const fileCipherIV = dataTxCipherResult.cipherIV;
 		const dataCipher = dataTxCipherResult.cipher ?? CIPHER_AES_256_GCM;
--- a/src/ardrive.ts
+++ b/src/ardrive.ts
@@ -28,7 +28,7 @@ import {
 import { ArFSDAO } from './arfs/arfsdao';
 import { IncrementalSyncOptions, IncrementalSyncResult, DriveSyncState } from './types/sync_types';
 import { CommunityOracle } from './community/community_oracle';
-import { deriveFileKey } from './utils/crypto';
+import { CIPHER_AES_256_CTR, CIPHER_AES_256_GCM, deriveFileKey } from './utils/crypto';
 import { ARDataPriceEstimator } from './pricing/ar_data_price_estimator';
 import {
 	FeeMultiple,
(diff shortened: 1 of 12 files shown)


vilenarios:
@coderabbitai review

vilenarios:
@coderabbitai review

